PR #2, opened 2019-11-07
sample slshrc file

gpwclark:
baked with some opinions.

### DIFF
--- a/contrib/slshrc
+++ b/contrib/slshrc
@@ -109,12 +109,12 @@
 		(str "")
 		(str "(" (str-trim branch) ")"))))
 
-#| prompt ensures
-	1. always a new line above prompt
-	2. if pushd has been used displays in reverse order dirs to popd
-	3. current path
-	4. if in git dir display current branch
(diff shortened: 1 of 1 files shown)


contrib/slshrc
```diff
@@ -0,0 +1,125 @@
...
+	(if (= branch "")
+		(str "")
+		(str "(" (str-trim branch) ")"))))
+
+#| prompt ensures
+	1. always a new line above prompt
+	2. if pushd has been used displays in reverse order dirs to popd
+	3. current path
+	4. if in git dir display current branch
+	5. cursor on near empty line |#
```
sstanfield: From what I have read the #| |# comments are bad form to use except when commenting out code...  I don't really care (and this is your config anyway) so you can use ; (or ;;) if you want or leave the block comment.  No idea why using block comments is considered bad form just read that multiple places.

gpwclark: nice, fixed it.

gpwclark:
looks good for now.

PR #11, opened 2019-12-15
more updates to slshrc

gpwclark:
- [la](https://github.com/sstanfield/slsh/compare/master...gpwclark:master#diff-79bccd575ff716c5469262f6810f359cR26)
    - would prefer this to be accessible via $la so I don't have to 'invoke' it with parens, but I'm not sure why the line errors with *last-command* not being found.
- [cd](https://github.com/sstanfield/slsh/compare/master...gpwclark:master#diff-79bccd575ff716c5469262f6810f359cR32)
    - this line blows the stack, is it possible to do
something like write over cd... I don't think it is? Alternatively, zsh automatically checks to see if a path is a directory and sort of obviates the need for cd. this might be a better solution than trying to alias a built-in. if there was some sort of hook that allowed us to check a command before evaluation... I could check if the input string was a directory and if so do a cd && ls of that directory. 
- [git completions](https://github.com/sstanfield/slsh/compare/master...gpwclark:master#diff-79bccd575ff716c5469262f6810f359cR214)
    - so my completion doesn't work... but we were brainstorming something else anyway... hooking into bash completions, which is ugly but until something like (shellac)[https://github.com/oilshell/oil/wiki/Shell-Autocompletion] is adopted idk what else we can do. So.. here's an idea: we write this [get_completions](https://brbsix.github.io/2015/11/29/accessing-tab-completion-programmatically-in-bash/) function to a tmp file, /tmp/bash-completions (or w/e we need to do to get that function in a bash environment), on startup and we could programmatically access all of bash's completions like ```/bin/bash -c "source /tmp/bash-completions &&  get_completions 'git checkout '"```


gpwclark:
ok, this is ready to go now. fixed la and cd, completions will be a different PR.

gpwclark:
ok, fixed the match statement.

gpwclark:
is this ready to go?

PR #12, opened 2019-12-31
Feature/bash completions

gpwclark:
* i can rebase off original PR to make diff cleaner once that's merged. the only code I'm talking about here is in the 
``` 
;; completions {{{  
...
;;;}}} 
``` 
section... fun vim folding thing

-  i spent way too much time trying to get a more slsh native version of this function and
   - completely failed to get it working for completions that required loading _completion_loader (no idea why)
   - introduced 2-3 bash sources which are ~1.5 hundredths of a second each which is pushing it
      - I used the bash builtin time fcn... we might should have one of those?)
      - https://input-delay.glitch.me/ to demo what keyboard lag can be like
- so, given you support multiline strings out of the box and bash -c doesn't barf when I hand it a string with a bash function inside, this seems to be the most elegant solution (where elegant means least ugly given what we're doing here).
- issue of, trimming the " mark off each of the elements in the list returned by the (get-bash-completion) fcn, ... I'll do that.
- larger issue of, when do we *not* call the completion loader... given that it will usually return something?
    - best idea I have thus far is ONLY using the completion loader if it does need to call _completion_loader?
    - i.e. in the get_completions fcn, simple bash functions like cd, don't need to load more complicated completion functions in `/usr/share/bash-completion/completions/`... perhaps this is the best proxy? if that makes sense?
- lastly, I can re-write the src copied from bash, it will not be as elegant as the original, but only reason to do so would be if licenses are incompatible (which they aren't) or if you don't like the way this works at all...
- i had trouble externalizing this functionality into  a separate .lisp file in the lisp dir so that's another reason I wanted feedback/help


sstanfield:
So I merged the other request and this one merged as well for some reason...  Need to figure out how I managed that for future merges...

For added a seperate file to the config directory did you try "load ~/.config/sl-sh/file-name.lisp?  I will test that later but that should be the simplest way to do it.

PR #18, opened 2020-01-13
update readme

gpwclark:
since this is very close to the top wanted a second pair of eyes. probably easiest if you "just edit".


sstanfield:
These updates are good and I think all accurate.  I should modify completions callback to use keywords now that sl-sh supports them but your description is accurate for the current system.  Thanks.

PR #21, opened 2020-02-02
primitive test runner

gpwclark:
so, similar to bash, sl-sh might need some cli flags so I can tell it to not run in "non interactive mode" and not search for an rc file etc.. Ideally this would be a sl-sh script but invoking sl-sh in sl-sh was... weird behavior. Just trying to get a dialogue going about this. I'd prefer the script to look like this: 
```
#!/bin/bash

for TEST in `ls tests | grep -Ev "^test.lisp$"`;
do
	echo "${TEST}"
	RET=$(sl-sh tests/${TEST} 2>/dev/null)
	if [[ ${RET} = "0" ]]; then
		echo "Pass"
	else
		echo "Fail"
	fi
done
```
but in sl-sh... but this doesn't work either, it never gets past the line `RET=$(sl-sh tests/${TEST} 2>/dev/null)`


sstanfield:
I agree that is should either automatically bypass the rc file when executing scripts or it should be an option.  The script you provided seems to need this change:
	sl-sh tests/${TEST} 2>/dev/null
	RET=$?
to get the actual return code.  If I do that it seems to work.  If that does not do it it for you let me know, you may be doing something in your rc to cause issues.

gpwclark:
@sstanfield now it's ready to go... finally

PR #31, opened 2020-05-21
AUR pre push hook

gpwclark:
so,

I'm interested to see what will happen currently before your pub key is in the aur... like, will, you, still be able to push to master? curious what the error is. 

In the middle of the file (after all the functions) there is this line:
```
90  if [[ ${BRANCH_NAME} != "master" && ${BRANCH_NAME} != "remotes/origin/master" ]]; then
  1 >···log "Not pushing master branch, not building docs or updating distributions."
  2 >···exit 0
  3 fi
```
uncomment out the exit and just do a bogus push to trick the check the hook.


arichnad:
Doesn't `.githooks/pre-push` run on the developers computer?

Does it make more sense to put this into a `.travis.yml` file?

PR #65, opened 2021-03-22
Feature/documentationier docs

gpwclark:


contrib/gpwclark/bash-completions.lisp
```diff
@@ -82,20 +83,18 @@
 	(args)
 	(let ((arg-str (str-cat-list " " args)))
 		((fn (completions)
+             (println "meow")
```
sstanfield, line 86: Think this was a mistake.

docs/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM jekyll/builder
```
sstanfield, line 1: Docker for building docs?  Awesome.

docs/Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
sstanfield, line 1: Github needs python and ruby?  Yikes.

lisp/core.lisp
```diff
@@ -517,4 +517,14 @@ Example:
      (new-item old-item lst &rest mods)
      `(nsubstitute! ,new-item ,old-item (collect-copy ,lst) ,@mods))
 
+(defn *identity*
```
sstanfield, line 520: Why the earmuffs?  We have never used them for functions before just other variables (otherwise then entire stdlib would be earmuffed).

lisp/lib.lisp
```diff
@@ -256,3 +256,89 @@ Example:
          (reducer (fn (fst nxt)
                      `((fn (_) (when _ ,fst)) ,nxt))))
         (iterator::reduce reducer first-rev rest-rev)))
+
+(defn mkli
```
sstanfield, line 260: Does this make more sense in shell.lisp?  Not a big deal either way so leave if you want to.

lisp/lib.lisp
```diff
@@ -256,3 +256,89 @@ Example:
...
+;;; *std-lib-syms-hash* and ns-auto-export must be the last two symbols defined
+;;; in the sl-sh standard library. In order to increase speed of ns-auto-export
+;;; a list of all symbols in the standard library is pre-computed by iterating
+;;; over each namespace in (ns-list), and calling it with (ns-symbols) and
+;;; adding that to a hash map, using it as a set to test for membership.
+;;; In order for this set of symbols to be complete the calls to (ns-list)
+;;; and (ns-symbols) must be done after all symbols and the namespaces they're
+;;; defined in have been called.
+
+(def *std-lib-syms-hash* (iterator::reduce
```
sstanfield, line 314: If these need to be the last thing then I think they should go in slsh-std.lisp after the last load (outside the wrapping get-error is probably fine but in it if there is any chance of error then in the get-error is also good).  I think lib.list should be loaded earlier, probably above shell.lisp since I think it is only a matter of time before shell.lisp will need to use something in lib.lisp.

sstanfield (commented):
I had a few comments, nothing big but a few of things that should be tweaked before merge.  I only skimmed all the html/css/etc stuff- I hope you did not have to generate that all by hand.

PR #78, opened 2021-05-05
Feature/new iters

gpwclark:


src/builtins_str.rs
```diff
@@ -787,6 +789,104 @@ fn builtin_char_is_whitespace(
...
+                    int_val += add;
+                }
+                if overflow {
+                    Err(LispError::new("error, overflow occurred adding unicode scalar values that compose provided grapheme interpreted as unsigned 32 byte integers: {:?}."))
+                } else {
+                    Ok(Expression::alloc_data(ExpEnum::Int(int_val as i64)))
+                }
+            }
+            _ => Err(LispError::new(
+                "function takes one grapheme, multiple were provided.",
```
sstanfield, line 824: In general (applies to all the errors you are raising) why did you not include the function name?  I know we talked about this once and yes the backtrace should probably take you to the call but including the name can just make things clearer (especially to not have to check your column in a long line of forms).  We can talk about this tonight though (maybe I am contradicting myself and we can run some tests but I think we will want the function names).

src/builtins_str.rs
```diff
@@ -787,6 +789,104 @@ fn builtin_char_is_whitespace(
...
+            strings.push(Expression::alloc_data(ExpEnum::String(
+                s.to_string().into(),
+                None,
+            )));
+        }
+        Ok(Expression::alloc_data(ExpEnum::Vector(strings)))
+    };
+    match &next_arg_d.data {
+        ExpEnum::String(s, _) => with_str(s),
+        ExpEnum::Char(s) => with_str(s),
```
sstanfield, line 861: Should probably also cover ExpEnum::CodePoint- is literally one unicode scalar.

src/builtins_str.rs
```diff
@@ -787,6 +789,104 @@ fn builtin_char_is_whitespace(
...
+        _ => Err(LispError::new(
+            "expected one argument of type Char or String",
+        )),
+    }
+}
+
+fn builtin_unicode_scalars(
+    environment: &mut Environment,
+    args: &mut dyn Iterator<Item = Expression>,
+) -> Result<Expression, LispError> {
```
sstanfield, line 840: So this cool.  Would it make more sense to return a list (vector) of CodePoints and then a way to format a codepoint (or just fix one of the print forms to do it if it does not).  Are you using this or just added for completeness?

src/builtins_str.rs
```diff
@@ -787,6 +789,104 @@ fn builtin_char_is_whitespace(
...
+    let next_arg = param_eval(environment, args, "unicode-scalars")?;
+    let next_arg_d = next_arg.get();
+    params_done(args, "unicode-scalars")?;
+    let with_str = |s: &str| {
+        let mut unicode_scalars: Vec<String> = Vec::new();
+        for c in s.chars() {
+            unicode_scalars.push(format!("{}", c.escape_unicode()));
```
sstanfield, line 847: I do not think this is an exact match for sl-sh unicode \u codes (would need to check).  We should use sl-sh encoding not rust if we are going to do this.

src/builtins_str.rs
```diff
@@ -787,6 +789,104 @@ fn builtin_char_is_whitespace(
...
+        _ => Err(LispError::new(
+            "expected one argument of type Char or String",
+        )),
+    }
+}
+
+fn builtin_str_hash(
+    environment: &mut Environment,
+    args: &mut dyn Iterator<Item = Expression>,
+) -> Result<Expression, LispError> {
```
sstanfield, line 871: Should this be str->float instead of str-hash?  Hashes are typically ints of some sort (I think) and we do not expose any hashing stuff in general (might be misleading). I like the f64 solution to the u/i64 issue.

sstanfield: Maybe str-hash is fine, we can talk about it.

src/builtins_system.rs
```diff
@@ -196,6 +197,22 @@ fn builtin_jobs(
     }
 }
 
+fn builtin_sys_time(
```
sstanfield, line 200: This should be 'builtin_epoch' following a consistent pattern between the sl-sh name and builtin name simplifies finding builtins later when you forget where they are.

lisp/shell.lisp
```diff
@@ -849,9 +849,10 @@ the temporary file are saved to the sl-sh history.
 Section: shell"
     ()
     (var fc-file (str (temp-dir) "/sl-sh-fc.txt"))
```
sstanfield, line 851: Maybe get rid of this var while your here.

PR #85, opened 2021-05-28
Feature/redo readme

gpwclark:
for your consideration


README.md
```diff
@@ -1,195 +1,105 @@
...
-```
-	(hash-set! *repl-settings* :vi_esc_sequence '("jk" 200))
-```
-The first argument is the list of chars that indicate the esc sequence and the
-seconds argument is the number of ms the readline library will wait to receive
-the full escape sequence.
+### 1. Get sl-sh
+- [Install git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
+    ```
+    git clone https://github.com/sstanfield/slsh
```
sstanfield, line 56: Need to point to the new org.

README.md
```diff
@@ -1,195 +1,105 @@
...
-```
-	(def tok-slsh-form-color shell::*fg-blue*)
-	(def tok-slsh-fcn-color shell::*fg-cyan*)
-	(def tok-default-color shell::*fg-default*)
-	(def tok-sys-command-color shell::*fg-white*)
-	(def tok-sys-alias-color shell::*fg-default*)
-	(def tok-string-color shell::*fg-magenta*)
-	(def tok-invalid-color shell::*fg-red*)
-```
+Either method will leave you with a binary target/release/sl-sh that will run the shell. The above docker command will produce a completely static binary while compiling with rust will be linked to you systems libc. You can use the musl target with cargo to produce a static binary with an installed rust.
```
sstanfield, line 73: Typo "your system's"?

README.md
```diff
@@ -1,195 +1,105 @@
...
-				cmd))
-
-	(defn __exec_hook (cmd-to-execute)
-		(let ((args-list (str-split " " cmd-to-execute)))
-				(match (length args-list)
-					(1 (change-dir-if-arg-is-dir (first args-list)))
-					(nil cmd-to-execute))))
-```
+Simple Lisp SHell (sl-sh) is a lisp based shell written in Rust. It is not POSIX
+compliant and makes no effort to be. It runs on the Linux and MacOS platforms.
```
sstanfield, line 30: Should work on any unix but it not regularly tested on others yet.  I may be setting up something to cover a BSD at some point though (a test running on a Pi at home for instance).

PR #86, opened 2021-06-01
fix broken link

arichnad:


arichnad:
never mind, looks like there is a PR for this already.

PR #92, opened 2022-01-05
Feature/umask

gpwclark:
adds builtin umask and eliminates use of libc crate directly and just relies on libc externed in nix when needed.


src/builtins_system.rs
```diff
@@ -979,4 +1348,242 @@ Example:
...
+
+When an integer is provided the value replaces the current value of the file creation mode mask.
+
+When a string is provided as a symbolic mode mask the current value of the system's umask is used
+to generate the new mask according to the specifications of the provided permissions operator. By
+default most Linux distros will set it to 022 or 002. If provided mode begins with a digit, it is
+interpreted as an octal number; if not, it is interpreted as a symbolic mode mask.
+
+
+You can set umask in your slshrc (for you) or /etc/profile (for all users).
```
sstanfield, line 1377: /etc/profile?

sstanfield (approved):
Really only had one doc specific comment.  Maybe the umask support code should go into its own file?  That is trivial though and I maybe wrong.  Using "str" as a variable name seems a little wrong since it is also a type but obviously it is valid syntax so not a problem (?).  Anyway this looks really good.

PR #94, opened 2022-04-28
Feature/regex and color

gpwclark:


sstanfield (commented):
This looks good, few minor things and a suggestion for a spin off PR (reader support for regex).

PR #97, opened 2022-07-10
Feature/proc macros

gpwclark:
I've finally got something working here. ~but it still needs some work~.

1. ~the main question I have rn is, what to do about some of the duplicate code that would be imported, e.g. the `ExpandVecToArgs` and `use std::convert::TryInto`~ these functions live in builtins_util
2. ~still on the fence about how the ExpandVecToArgs stuff works generally~ found a work around with TryIntoExpression that is a bit verbose but... ultimately fine.
3. ~i need to figure out what the testing story is like~ the ./run-tests.lisp works fine!
4. ~need to explore handling variadic arguments because that's possible in rust and sl-sh and will need some sort of extra code~.... this may be out of scope we'll see as it evolves
5. to keep us from getting into runtime errors, i've included this static_assertions dep and I make sure the appropriate implementations of Into/TryInto exist at compile time to help us figure out what we do need to implement as we convert builtin methods to use this.


src/builtins_types.rs
```diff
@@ -435,32 +436,33 @@ fn builtin_str_to_float(
...
+/// Usage: (int->float int) -> float
+///     Cast an int as a float.
+/// Section: type
+/// Example:
+///     (test::assert-equal 0 (int->float 0))
+///     (test::assert-equal 10 (int->float 10))
+///     (test::assert-equal -101 (int->float -101))
+///     (test::assert-error (int->float "not int"))
+#[sl_sh_fn(fn_name = "int->float")]
+fn int_to_float(int: i64) -> f64 {
```
sstanfield, line 448: I should look at the macro but being lazy.  If this could cause an error then can you return a Result here and have that become a Lisp error?  Out of scope for this PR but we should review these number routines for precision- it might make more sense to use verify the conversions are valid and error out of not (I think there is a TryFrom for these).  This one may always be safe but f64 to i64 might not be.

sstanfield (approved):

src/builtins_types.rs
gpwclark: addressed in the other PR

PR #98, opened 2022-07-17
Feature/proc macros types

gpwclark:
- support returning LispResult types on functions with proc_macro_attribute
- support eval_values attribute

### DIFF
--- a/src/builtins_types.rs
+++ b/src/builtins_types.rs
@@ -9,395 +9,501 @@ use crate::eval::*;
 use crate::interner::*;
 use crate::types::*;
 
-fn builtin_type(
-    environment: &mut Environment,
-    args: &mut dyn Iterator<Item = Expression>,
-) -> Result<Expression, LispError> {
-    if let Some(arg) = args.next() {
-        if args.next().is_none() {
-            let arg = eval(environment, arg)?;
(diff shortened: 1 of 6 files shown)


sstanfield (commented):
I think you are on the right track with eval_no_values.  Unless you add support for values, which on first blush would be way to complex for something that would almost never be used, then you want eval_no_values when using your macro.  Something like values? that needs the actual values will not work, but I think cases like that will be few and far between so maybe just leave that one as a "native" builtin vs using your macro (I think that is perfectly reasonable vs making your macro more complex for a tiny handful of buitins with odd requirements).  You could add an attribute to tell the macro NOT to eval the expression before passing it in for special cases like this (and special forms).  Note that will not really work with slosh since it compiles things instead of evals them- so might not be worth bothering with.

src/builtins_types.rs
```diff
@@ -520,7 +514,12 @@ fn int_to_float(int: i64) -> f64 {
 /// (test::assert-error (float->int "not int"))
 #[sl_sh_fn(fn_name = "float->int")]
 fn float_to_int(float: f64) -> i64 {
-    float as i64
```
gpwclark, line 465: here is the float to int logic

sstanfield (approved):

PR #102, opened 2022-11-10
Feature/experiment inlining

gpwclark:
- some tweaks to the sl-sh codebase in lieu of the updates to the proc macro library.
- removes some dead test code and moved a few of the salvageable macro test cases over to builtin_types.rs where they were more at home
- update clippy


gpwclark:
Uses new dep that does inlining. Deleted some old test code that helped me verify what the code I was generating should look like.

PR #112, opened 2024-01-17
Test rust docs in rust

gpwclark:
- made a test that checks the format of rust docs, or at least, the strings the get-prop function gets for the :doc-string keywork
- migrated some docs over from sl-sh
- added some EXEMPTIONS

- did a bit of work with the ErrorStrings I'm basing my errors in the conversion lib functions.

main point was to get the docstrings in rust and make sure they were formatted correctly so if you break something in rust, cargo test tells you. obviously incomplete because of the EXEMPTIONS set but it is a start. Mainly a pre-cursor to me actually eval-ing the Example: sections for the docs that do have them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


sstanfield (approved):
This looks good.  Thanks for filling out a bunch of the doc strings for the compiler builtins.

PR #113, opened 2024-01-18
Some fixes/improvements to GC.

sstanfield:
This gives piars/concells and values there own storage area in the heap which reduces allocations (these are simple Copy types that can live directly in a Vec) and should speed up using those types.  Also fixes some big holes in GC that would have leaked continutations and call frames...

Need to spend some time on more comprehensive tests of the heap in isolation.  Now that there are multiple "storage" areas testing across types in different areas and probably other things should be included.

### DIFF
--- a/vm/src/heap.rs
+++ b/vm/src/heap.rs
@@ -51,8 +51,6 @@ enum Object {
     Vector(Arc<Vec<Value>>),
     Map(Arc<HashMap<Value, Value>>),
     Bytes(Arc<Vec<u8>>),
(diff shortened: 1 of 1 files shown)


gpwclark (approved):
nice. got to read up on mark/sweep and tri-color marking to figure out what this "grey" business is. Looks good as far as I can tell. Def would like to see the test suite grow here but this is a good step forward as it looks like you're doing some important accounting now that wasn't done before. good catch.

PR #118, opened 2024-02-10
Introduce let-while special form.

sstanfield:
Getting this on github, it has been languishing for a week or so and want to track it.

This is really a WIP, needs some docs and bit of cleanup.


gpwclark (approved):
definitely still trying to wrap my head around a lot of the vm internals. 

definitely definitely trying to make the slosh test script a reality so I can help give you some guarantees about how new slosh forms are functioning.

always a fan of more docs.

PR #122, opened 2024-02-23
2. make an integration test that runs a slosh script Closes #111

### LINKED ISSUE #111: slosh shell tests
slosh needs tests for shell like things, i.e.

1. create some random directory hierarchies and text files
2. test different redirect/pipes 
3. suspend/kill/job control
anything to catch a regression we might run into on the shell,

shell-tests/test-simple-redirect.slosh
shell-tests/test-simpe-pipe.losh

etc. and a test runner to test them all.

gpwclark:
Issue #111: formalize slosh script integration test running, all slosh scripts in slosh/test/*.slosh get run, those that end in *_fail.slosh should return non-zero exit codes


sstanfield (approved):
Nice refactors.  This will conflict with my -c PR, you can merge first (which will make that PR even smaller) or absorb that PR if you want since your refactors make it simpler.

PR #123, opened 2024-02-23
feat: support for F56 56-bit floating point numbers

StevenLove:
### Overview
This implementation allows us to store floats with greater precision.
The existing implementation stores `slvm::Value::Float` as a wrapped `f32` and ignores the other 3 available bytes.

### Changes
- `builtins/src/types/numbers.rs` primarily just has the new test cases (100 of them!)
- `vm/src/value.rs` has the actual `F56` implementation and some changes from the previous implementation using `F32Wrap`
- `vm/src/vm/macros.rs` might require some additional attention. The `get_float` macro here is now set to operate on `f64` types. But a similar `get_float` macro in `value.rs` is set to `f32`.  These implications should be explored.

### DIFF
--- a/builtins/src/types/numbers.rs
+++ b/builtins/src/types/numbers.rs
@@ -116,9 +116,13 @@ mod tests {
     }
 
     #[test]
+    /// Declares a bunch of f64s meant to represent a wide range of values of various edge cases
+    /// Converts each f64 to f56 and back to f64 to see how it changes
+    /// If they are equal, then that conversion was successful
+    /// If they are unequal, then it may still be successful if the f64 was outside of the range that the F56 could represent
+    /// A debug function is provided to visually inspect the bytes of the f64, F56, and corresponding f32
+
     fn test_f56() {
-        /* For each of these test numbers, convert them from f64 to f56 and back and
-        verify that the number hasn't changed much */
         let numbers_to_test = [
             f64::from_bits(0x0000_0000_0000_0000u64),
             f64::from_bits(0x0000_0000_0000_0001u64),
@@ -247,32 +251,32 @@ mod tests {
 
         for index in 0..numbers_to_test.len() {
             let orig_f64 = numbers_to_test[index];
-            /* Calculate the values  */
+            // Calculate the values
             let f32 = orig_f64 as f32;
             let f56 = F56::from(orig_f64);
             let back_to_f64: f64 = f56.into();
             let f32_diff = ((f32 as f64) - orig_f64).abs();
             let f56_diff = (back_to_f64 - orig_f64).abs();
 
-            /* If the converted value equals the original, we passed the test */
+            // If the converted value equals the original, we passed the test
             if back_to_f64 == orig_f64 {
                 continue;
             }
 
-            /* Signs must match */
+            // Signs must match
             if orig_f64.is_sign_positive() != back_to_f64.is_sign_positive() {
                 debug(orig_f64, index);
                 panic!("Signs don't match");
             }
 
-            /* NaNs must match */
+            // NaNs must match
             if orig_f64.is_nan() != back_to_f64.is_nan() {
                 debug(orig_f64, index);
                 panic!("original f64 and f56 should either both be NaN or neither be NaN");
             }
 
-            /* Both must be finite or infinite
-            Unless the f64 is very large and the F56 is infinite */
+            // Both must be finite or infinite
+            // Unless the f64 is very large and the F56 is infinite
             if orig_f64.is_infinite() != back_to_f64.is_infinite() {
                 if orig_f64.is_infinite() {
                     debug(orig_f64, index);
@@ -286,9 +290,8 @@ mod tests {
                 }
                 continue; // Don't check magnitude of difference if one is infinite
             }
-            /* Both must be zero or nonzero
-            Unless the f64 is very small and the F56 is zero
-             */
+            // Both must be zero or nonzero
+            // Unless the f64 is very small and the F56 is zero
             if (orig_f64 == 0.0) != (back_to_f64 == 0.0) {
                 if orig_f64 == 0.0 {
                     debug(orig_f64, index);
@@ -303,14 +306,14 @@ mod tests {
                 continue;
             }
 
-            /* The absolute difference must be small
-            smaller than the difference between an f32 and the f64 */
+            // The absolute difference must be small
+            // smaller than the difference between an f32 and the f64
             if f56_diff > f32_diff {
                 debug(orig_f64, index);
                 panic!("f56 is less accurate than the f32 conversion");
             }
 
-            /* The relative difference must be less than F56's EPSILON */
+            // The relative difference must be less than F56's EPSILON
             let relative_difference = f56_diff / orig_f64.abs();
             if relative_difference > F56::EPSILON {
                 debug(orig_f64, index);
--- a/builtins/src/types/numbers.rs
+++ b/builtins/src/types/numbers.rs
@@ -41,7 +41,60 @@ mod tests {
     use crate::types::SlFrom;
     use crate::types::SlInto;
     use compile_state::state::new_slosh_vm;
+    use core::panic;
+    use slvm::F56;
     use slvm::{to_i56, Value};
+
+    pub fn log_f32(f: f32) -> String {
+        let bytes = f.to_be_bytes();
+        let word = u32::from_be_bytes([bytes[0], bytes[1], bytes[2], bytes[3]]);
+        let f32_biased_exponent: u8 = (word >> 23) as u8; // first 8 bits after the sign bit
+        let mut true_exponent: i16 = (f32_biased_exponent as i16) - 127; // remove the bias of 2^7-1
+        if f32_biased_exponent == 0 {
+            true_exponent = 0;
+        }
+        let f32_mantissa = word & 0x007f_ffff; // everything after first 9 bits
+                                               // print the f32 in scientific notation, the true exponent in decimal, and the mantissa in decimal, and the hex word
+        format!(
+            "f32: {:.5e}, true exponent: {}, mantissa: {:016x}, word: {:016x}",
+            f, true_exponent, f32_mantissa, word
+        )
+    }
+    pub fn log_f56(f: F56) -> String {
+        let bytes7 = f.0;
+        let word = u64::from_be_bytes([
+            0, bytes7[0], bytes7[1], bytes7[2], bytes7[3], bytes7[4], bytes7[5], bytes7[6],
+        ]);
+        let f56_biased_exponent: u16 = ((word >> 45) as u16) & 0x3FF; // first 10 bits after the sign bit
+        let f56_mantissa = word & 0x1FFFF_FFFF_FFFF; // everything after first 10 bits
+        let mut true_exponent = f56_biased_exponent as i16 - 511; // remove the bias of 2^9-1
+        if f56_biased_exponent == 0 {
+            true_exponent = 0;
+        }
+        format!(
+            "f56: {:.5e}, true exponent: {}, mantissa: {:016x}, word: {:016x}",
+            f64::from(f),
+            true_exponent,
+            f56_mantissa,
+            word
+        )
+    }
+    pub fn log_f64(f: f64) -> String {
+        let bytes = f.to_be_bytes();
+        let word = u64::from_be_bytes([
+            bytes[0], bytes[1], bytes[2], bytes[3], bytes[4], bytes[5], bytes[6], bytes[7],
+        ]);
+        let f64_biased_exponent: i16 = ((word & 0x7ff0_0000_0000_0000) >> 52) as i16; // first 11 bits after the sign bit
+        let mut true_exponent: i16 = f64_biased_exponent - 1023; // remove the bias of 2^10-1
+        if f64_biased_exponent == 0 {
+            true_exponent = 0;
+        }
+        let f64_mantissa = word & 0x001f_ffff_ffff_ffff; // everything after first 11 bits
+        format!(
+            "f64: {:.5e}, true exponent: {}, mantissa: {:016x}, word: {:016x}",
+            f, true_exponent, f64_mantissa, word
+        )
+    }
 
     #[test]
     fn test_i32_conversions_rust_to_value() {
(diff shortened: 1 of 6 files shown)


vm/src/value.rs
```diff
@@ -84,24 +84,226 @@ impl<'vm, ENV> Iterator for PairIter<'vm, ENV> {
...
+ * maximum finite value is 1.34e154
+*/
+#[derive(Copy, Clone)]
+pub struct F56(pub [u8; 7]);
+impl Eq for F56 {}
+impl PartialEq for F56 {
     fn eq(&self, other: &Self) -> bool {
-        (self.0 - other.0).abs() < f32::EPSILON
-        //self.0.to_bits() == other.0.to_bits()
+        f64::from(*self) == f64::from(*other)
```
sstanfield, line 114: I think you want to keep the original "< EPSILON" check here otherwise Eq and PartialEq may not be usable.  Implementing these is kinda sloppy at all but I want to be able to hash a Value...

sstanfield (commented):
This looks great, thanks.  I want to play with it some with an eye toward making sure it does not tank math performance (I am not expecting it to) before merge.  For reference this is a dumb benchmarking script I use and it does stuff with FP so may actually be valid in this case:
```
(defn eval-pol (n x)
  (let (su 0.0
        mu 10.0
        pu 0.0
        pol (make-vec 100 0.0))
    (dotimes-i i n
        (set! su 0.0)
        (dotimes-i j 100
             (set! mu (/ (+ mu 2.0) 2.0))
             (set! pol.~j mu))
        (dotimes-i j 100
          (set! su (+ pol.~j (* su x))))
            (set! pu (+ pu su)))
    (prn pu)))
```
You can load it and then I usually do (eval-pol 500000 .2) - make sure to use a release build.  It should take a few seconds to run, it does a lot of useless work and produces 1250000.  Having an impact on performance is not a problem I just don't want a huge impact (which can probably be fixed if it is there).  We also need to start getting proper benchmarks, etc setup but this stupid script has been useful for letting me know when something in the bytecode loop trashed performance, etc.

builtins/src/types/numbers.rs
```diff
@@ -61,4 +114,212 @@ mod tests {
         let val = to_i56(7_i32 as i64);
         let _val: i32 = i32::sl_from(&val, vm).expect("Value can be converted to i32");
     }
+
+    #[test]
+    fn test_f56() {
+        /* For each of these test numbers, convert them from f64 to f56 and back and
```
gpwclark, line 120: in rust functions are documented like this:
```
/// a documentation comment has 3 slashes
/// other fun stuff
fn myfun(f: f64)  {
...
}
```

and for no real good reason, the /* ... */ style comment isn't really used in Rust: https://doc.rust-lang.org/book/ch03-04-comments.html#:~:text=In%20Rust%2C%20the%20idiomatic%20comment,%23!%5B

the compiler seems to support it but we don't use it... for some reason. Because of `cargo fmt` ... I think, Rust devs are a bit more religious about this. EIther way in this context this comment should just be a `///` doc comment above the function because it's useful and it describes what the test does.

builtins/src/types/numbers.rs
```diff
@@ -61,4 +114,212 @@ mod tests {
...
+            let f56_relative_difference = f56_diff / orig_f64.abs();
+            if f56_relative_difference > 0.0 {
+                println!("f56 relative difference {:.5e}", f56_relative_difference);
+            }
+            println!("")
+        }
+
+        for index in 0..numbers_to_test.len() {
+            let orig_f64 = numbers_to_test[index];
+            /* Calculate the values  */
```
gpwclark, line 250: for comments like these within functions, use the double slash... even if they're multiline, which I know is annoying.
```
// this comment is about assiging foo
// and can extend multiple lines.
let foo = 0.0;
```

vm/src/value.rs
```diff
@@ -84,24 +84,226 @@ impl<'vm, ENV> Iterator for PairIter<'vm, ENV> {
...
+ * Exponent bits range from 0 to 1023
+ * they represent -511 to +512 but are stored biased by +511
+ * the exponent of -511 is reserved for the number 0 and subnormal numbers
+ * the exponent of +512 is reserved for infinity and NaN
+ * so normal exponents range from -510 to +511
+ *
+ * smallest positive subnormal value is 8.48e-168 (2^-555)
+ * smallest positive normal value is 2.98e-154 (2^-510)
+ * maximum finite value is 1.34e154
+*/
```
gpwclark, line 108: turn this into a `///` doc comment. Documnetation comments are additionally useful because
1. They show up in the docs when you build them with `cargo docs`, rust builds documentation for crates automatically.
2. They can have tests: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

builtins/src/types/numbers.rs
StevenLove: Thanks for the info! It's funny what you miss when you learn a language through its linter. I'll update with proper comments

gpwclark:
> This looks great, thanks. I want to play with it some with an eye toward making sure it does not tank math performance (I am not expecting it to) before merge. For reference this is a dumb benchmarking script I use and it does stuff with FP so may actually be valid in this case:
> 
> ```
>   (let (su 0.0
>         mu 10.0
>         pu 0.0
>         pol (make-vec 100 0.0))
>     (dotimes-i i n
>         (set! su 0.0)
>         (dotimes-i j 100
>              (set! mu (/ (+ mu 2.0) 2.0))
>              (set! pol.~j mu))
>         (dotimes-i j 100
>           (set! su (+ pol.~j (* su x))))
>             (set! pu (+ pu su)))
>     (prn pu)))
> ```
> 
> You can load it and then I usually do (eval-pol 500000 .2) - make sure to use a release build. It should take a few seconds to run, it does a lot of useless work and produces 1250000. Having an impact on performance is not a problem I just don't want a huge impact (which can probably be fixed if it is there). We also need to start getting proper benchmarks, etc setup but this stupid script has been useful for letting me know when something in the bytecode loop trashed performance, etc.

Steve this script fails to compile because the `x` on this line is undefined
```
           (set! su (+ pol.~j (* su x))))
```

sstanfield:
> > This looks great, thanks. I want to play with it some with an eye toward making sure it does not tank math performance (I am not expecting it to) before merge. For reference this is a dumb benchmarking script I use and it does stuff with FP so may actually be valid in this case:
> > ```
> >   (let (su 0.0
> >         mu 10.0
> >         pu 0.0
> >         pol (make-vec 100 0.0))
> >     (dotimes-i i n
> >         (set! su 0.0)
> >         (dotimes-i j 100
> >              (set! mu (/ (+ mu 2.0) 2.0))
> >              (set! pol.~j mu))
> >         (dotimes-i j 100
> >           (set! su (+ pol.~j (* su x))))
> >             (set! pu (+ pu su)))
> >     (prn pu)))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can load it and then I usually do (eval-pol 500000 .2) - make sure to use a release build. It should take a few seconds to run, it does a lot of useless work and produces 1250000. Having an impact on performance is not a problem I just don't want a huge impact (which can probably be fixed if it is there). We also need to start getting proper benchmarks, etc setup but this stupid script has been useful for letting me know when something in the bytecode loop trashed performance, etc.
> 
> @sstanfield this script fails to compile because the `x` on this line is undefined
> 
> ```
>            (set! su (+ pol.~j (* su x))))
> ```
> 
> also if we're benchmarking it i'll probably remove the prn.

@gpwclark Well that is just embarrassing.  I did load this and test it to avoid this issue but that is an obvious error and there are more than one versions of this script so I probably loaded the wrong one.  I will sort it and commit a version of this to the repo so we can stay in sync.  As far as the prn goes it outside all the loops and I like to print the result because sometimes I break stuff and this becomes a redflag (although there is a unit test based around this benchmark but I think it uses raw byte code and a lot less loops so not exactly the same).  There is also an example that does the same thing but using raw byte code, it is good for seeing just how unoptimized the computer is (I think it is currently 2-3 times faster)...  This all pretty weak benchmarking but I have found it useful for making sure I don't accidentally trash performance when doing something new (since being faster was one of the reasons slim even exists...)

vm/src/value.rs
StevenLove: Good point. Would it be reasonable to keep F56 acting the same as the primitives and instead add this epsilon diff test elsewhere like in the hash fn? I see some value in making F56 act like the primitives but not if it overcomplicates everything else.

Further, the hash function looks like it converts the f56 to an f64 and then to bits to write a u64. Doesn't this mean that two f56s within EPSILON of each other would still have different bit patterns and hash to different values? I'm not familiar with all the hash so forgive me if I'm going totally the wrong direction with this.

vm/src/value.rs
```diff
@@ -270,7 +472,7 @@ impl Value {
         match &self {
             Value::Byte(b) => Ok(*b as f32),
             Value::Int(i) => Ok(from_i56(i) as f32),
-            Value::Float(f) => Ok(f.0),
+            Value::Float(f) => Ok(f32::from(*f)),
```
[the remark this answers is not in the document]
StevenLove: I used f32 here because `cargo test` was not passing the `test_pol` function with f64. I'm realizing now what the issue is and it has wider implications. Here's the test result:
```
---- vm::tests::test_pol stdout ----
thread 'vm::tests::test_pol' panicked at vm/src/vm.rs:914:9:
assertion `left == right` failed
  left: 12499.999999998836
 right: 12500.0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
The F56 only intends to have 12-14 digits of decimal precision and this 12499.999999998836 result actually has about that many good digits, but because we convert the F56 to an f64, the function that converts an f64 to a string expects a few extra digits of precision that the F56 just didn't have. So pretty much anywhere convert F56 to f64, the resulting f64 won't properly convert to a decimal string. The same thing happens if we convert an f32 to an f64
```
        let v: f32 = 4.8;
        println!("{:?}", v); // 4.8
        println!("{:?}", v as f64); // 4.800000190734863
```

The proper solution depends on the context of when / how often we convert from Value back to primitives. We could always round before we convert to f64, but I'd like to avoid that if possible. If the string representation only matters for our tests then we can just convert to f32 to check the string. Let me know what you think.

Somewhat related, I already implemented a custom Display trait for F56 that only relies on the 12 decimal digits of precision because of the `test_tok_floats` test in `reader.rs`.

sstanfield:
> > > This looks great, thanks. I want to play with it some with an eye toward making sure it does not tank math performance (I am not expecting it to) before merge. For reference this is a dumb benchmarking script I use and it does stuff with FP so may actually be valid in this case:
> > > ```
> > >   (let (su 0.0
> > >         mu 10.0
> > >         pu 0.0
> > >         pol (make-vec 100 0.0))
> > >     (dotimes-i i n
> > >         (set! su 0.0)
> > >         (dotimes-i j 100
> > >              (set! mu (/ (+ mu 2.0) 2.0))
> > >              (set! pol.~j mu))
> > >         (dotimes-i j 100
> > >           (set! su (+ pol.~j (* su x))))
> > >             (set! pu (+ pu su)))
> > >     (prn pu)))
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You can load it and then I usually do (eval-pol 500000 .2) - make sure to use a release build. It should take a few seconds to run, it does a lot of useless work and produces 1250000. Having an impact on performance is not a problem I just don't want a huge impact (which can probably be fixed if it is there). We also need to start getting proper benchmarks, etc setup but this stupid script has been useful for letting me know when something in the bytecode loop trashed performance, etc.
> > 
> > 
> > @sstanfield this script fails to compile because the `x` on this line is undefined
> > ```
> >            (set! su (+ pol.~j (* su x))))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > also if we're benchmarking it i'll probably remove the prn.
> 
> @gpwclark Well that is just embarrassing. I did load this and test it to avoid this issue but that is an obvious error and there are more than one versions of this script so I probably loaded the wrong one. I will sort it and commit a version of this to the repo so we can stay in sync. As far as the prn goes it outside all the loops and I like to print the result because sometimes I break stuff and this becomes a redflag (although there is a unit test based around this benchmark but I think it uses raw byte code and a lot less loops so not exactly the same). There is also an example that does the same thing but using raw byte code, it is good for seeing just how unoptimized the computer is (I think it is currently 2-3 times faster)... This all pretty weak benchmarking but I have found it useful for making sure I don't accidentally trash performance when doing something new (since being faster was one of the reasons slim even exists...)

@gpwclark So I truncated that lisp snippet, x is a parameter.  The correct snippet is:
```
(defn eval-pol (n x)
  (let (su 0.0
        mu 10.0
        pu 0.0
        pol (make-vec 100 0.0))
    (dotimes-i i n
        (set! su 0.0)
        (dotimes-i j 100
             (set! mu (/ (+ mu 2.0) 2.0))
             (set! pol.~j mu))
        (dotimes-i j 100
          (set! su (+ pol.~j (* su x))))
            (set! pu (+ pu su)))
    (prn pu)))
```
That will fix it by defining the function eval-pol...

gpwclark:
> > > > This looks great, thanks. I want to play with it some with an eye toward making sure it does not tank math performance (I am not expecting it to) before merge. For reference this is a dumb benchmarking script I use and it does stuff with FP so may actually be valid in this case:
> > > > ```
> > > >   (let (su 0.0
> > > >         mu 10.0
> > > >         pu 0.0
> > > >         pol (make-vec 100 0.0))
> > > >     (dotimes-i i n
> > > >         (set! su 0.0)
> > > >         (dotimes-i j 100
> > > >              (set! mu (/ (+ mu 2.0) 2.0))
> > > >              (set! pol.~j mu))
> > > >         (dotimes-i j 100
> > > >           (set! su (+ pol.~j (* su x))))
> > > >             (set! pu (+ pu su)))
> > > >     (prn pu)))
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > You can load it and then I usually do (eval-pol 500000 .2) - make sure to use a release build. It should take a few seconds to run, it does a lot of useless work and produces 1250000. Having an impact on performance is not a problem I just don't want a huge impact (which can probably be fixed if it is there). We also need to start getting proper benchmarks, etc setup but this stupid script has been useful for letting me know when something in the bytecode loop trashed performance, etc.
> > > 
> > > 
> > > @sstanfield this script fails to compile because the `x` on this line is undefined
> > > ```
> > >            (set! su (+ pol.~j (* su x))))
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > also if we're benchmarking it i'll probably remove the prn.
> > 
> > 
> > @gpwclark Well that is just embarrassing. I did load this and test it to avoid this issue but that is an obvious error and there are more than one versions of this script so I probably loaded the wrong one. I will sort it and commit a version of this to the repo so we can stay in sync. As far as the prn goes it outside all the loops and I like to print the result because sometimes I break stuff and this becomes a redflag (although there is a unit test based around this benchmark but I think it uses raw byte code and a lot less loops so not exactly the same). There is also an example that does the same thing but using raw byte code, it is good for seeing just how unoptimized the computer is (I think it is currently 2-3 times faster)... This all pretty weak benchmarking but I have found it useful for making sure I don't accidentally trash performance when doing something new (since being faster was one of the reasons slim even exists...)
> 
> @gpwclark So I truncated that lisp snippet, x is a parameter. The correct snippet is:
> 
> ```
> (defn eval-pol (n x)
>   (let (su 0.0
>         mu 10.0
>         pu 0.0
>         pol (make-vec 100 0.0))
>     (dotimes-i i n
>         (set! su 0.0)
>         (dotimes-i j 100
>              (set! mu (/ (+ mu 2.0) 2.0))
>              (set! pol.~j mu))
>         (dotimes-i j 100
>           (set! su (+ pol.~j (* su x))))
>             (set! pu (+ pu su)))
>     (prn pu)))
> ```
> 
> That will fix it by defining the function eval-pol...

hmm nice. I'm going to try and use this with bencher. but that would move the iteration out of lisp.

vm/src/value.rs
StevenLove: Thanks for explaining the conundrum. I now understand why floats are throwing a wrench in the goal to make any Value a proper hash key. There's more I need to understand about the overall project direction in order to make judgment on which imperfect solution we should go forward with. 

In the meantime, I think it would be suitable to implement partial equality for F56 to be 'within EPSILON' and to allow NaN == NaN to be true so we can have the Eq trait. This deviation from primitive float behavior is unfortunate, but the capitalization of F56 in contrast to f64 should help remind everyone that it isn't a primitive. And arguably the behavior is more intuitive than the behavior of primitive floats.

In an effort to merge this PR, I'd like to implement this option and then open a github issue where we can further discuss whether to make F56 and Value implement Eq and Hash, and how to implement PartialEq for F56, and when/whether to panic or throw errors if someone tries to use a float as a hash key.

vm/src/value.rs
```diff
@@ -84,24 +84,238 @@ impl<'vm, ENV> Iterator for PairIter<'vm, ENV> {
...
+/// maximum finite value is 1.34e154
+#[derive(Copy, Clone)]
+pub struct F56(pub [u8; 7]);
+impl Eq for F56 {}
+impl PartialEq for F56 {
     fn eq(&self, other: &Self) -> bool {
-        (self.0 - other.0).abs() < f32::EPSILON
-        //self.0.to_bits() == other.0.to_bits()
+        // Allow NaN == NaN so equality is reflexive and we can impl Eq to use F56 as a hash key
+        if f64::from(*self).is_nan() && f64::from(*other).is_nan() {
```
sstanfield, line 111: You are calling f64::from() twice, once here and again below.  Recommend you move the let value and let other_value statements up and only call the from once per value vs twice.  The from is not super costly but it is not free.

vm/src/value.rs
sstanfield: This is reasonable and is a strict improvement over over the old code as well.  Opening an issue is a good idea to hash this out but it is out of scope for this PR.  This is a scripting language so being a bit looser than Rust is probably the right move in most cases (thing Go for things like this).

sstanfield (approved):
I checked the performance, and very roughly the heavy float loop took a 30ish% hit.  I think that is OK for this stage of things.  There is most likely room to optimize and most/all of it will be in the From/Into f64 traits.  That will probably be a PITA and may be a case of optimizing for one platform being suboptimal on another but you may want to spend some time later playing with those traits and see if you can close the gap some more.

Approving unless you make any more changes (please clear this if you make substantial changes).  @gpwclark You can merge when you are ready to approve. 

Thanks

vm/src/value.rs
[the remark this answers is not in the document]
StevenLove: Bump. Any thoughts on this?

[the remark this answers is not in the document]
sstanfield: It would be good to not have the weird precision stuff show up like this.  If I understand this correctly then if we use the Display on F56 directly (ie don't convert to a f64 first) then it will work fine?  If so I think we should make sure that the lisp code that displays floats should do this.  For tests it is probably fine to convert to f32 as long as that won't loose anything we need otherwise we need to probably use Eq on F56 that I think takes care of this with epsilon now?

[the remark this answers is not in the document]
StevenLove: Absolutely correct, F56's Display should work fine. And for displaying the lisp values, the 'display_value' function in value.rs looks to appropriately use the F56 Display.

gpwclark:
> > This looks great, thanks. I want to play with it some with an eye toward making sure it does not tank math performance (I am not expecting it to) before merge. For reference this is a dumb benchmarking script I use and it does stuff with FP so may actually be valid in this case:
> > ```
> >   (let (su 0.0
> >         mu 10.0
> >         pu 0.0
> >         pol (make-vec 100 0.0))
> >     (dotimes-i i n
> >         (set! su 0.0)
> >         (dotimes-i j 100
> >              (set! mu (/ (+ mu 2.0) 2.0))
> >              (set! pol.~j mu))
> >         (dotimes-i j 100
> >           (set! su (+ pol.~j (* su x))))
> >             (set! pu (+ pu su)))
> >     (prn pu)))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can load it and then I usually do (eval-pol 500000 .2) - make sure to use a release build. It should take a few seconds to run, it does a lot of useless work and produces 1250000. Having an impact on performance is not a problem I just don't want a huge impact (which can probably be fixed if it is there). We also need to start getting proper benchmarks, etc setup but this stupid script has been useful for letting me know when something in the bytecode loop trashed performance, etc.
> 
> @sstanfield this script fails to compile because the `x` on this line is undefined
> 
> ```
>            (set! su (+ pol.~j (* su x))))
> ```
> 
> also if we're benchmarking it i'll probably remove the prn.

ok I merged the bench marking stuff. We now have before and after benchmarks in bencher.dev with the new float benchmarks. Here is the before:
https://github.com/sl-sh-dev/sl-sh/pull/127
and I opened a sample PR to trigger the GH action to run with the new code. I took steven's master branch, rebased it on the new master with bencher feature, and updated the float code in the benchmark for f56 as a PoC to measure impact. Loooks good:
https://github.com/sl-sh-dev/sl-sh/pull/128

@StevenLove update your PR against master you're going to need the commit 
```
commit 1edc53ca0d2617dcaa0dc303f0b8d5ef97c5a8db (HEAD -> test-new-floats, origin/test-new-floats)
Author: gpwclark <gpwclark@gmail.com>
Date:   Wed Feb 28 00:22:35 2024 -0500

    update for bencher

``` 

from the branch test-new-floats when you update. We should merge your PR (I'll close the one I opened) but I had to open one to trigger the GH action to run with the proper credentials to submit the data to bencher.dev. Anyway, I say merge! Also can you see these:

https://bencher.dev/perf/sl-sh?key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&report=42268bc0-ffe3-42a3-9a64-a430b25e7b92&branches=98180d27-2dbf-4981-90ca-874221ddf395&testbeds=2e20891d-0b52-4c2d-b14e-7937d1311d1a&benchmarks=17622464-e11b-4753-85a2-9d7e0c3d14ce%2C9722a538-12ee-4a63-b438-5cb6c62010a0%2Cc98a610b-f03a-491a-8afd-fffb76e64e9e&measures=cb6dca16-7098-4385-8bee-be93c61aea7a&start_time=1706505783000&end_time=1709097903000&clear=true

gpwclark (approved):

PR #126, opened 2024-02-26
Run slosh tests

gpwclark:
- all the docs written in rust are  now known to be either exempt (and should be fixed in due course) or have properly formatted docs.
- made the test assert builtins, beefed up equal?, added some shims for ns-import (noop), and alias for test::assert-equal, test::assert-true, etc to preserve existing lisp code.
- added some docs
- added the slosh docs test that runs when the lisp-test feature flag is enabled
- have skeleton for tests the evaluate lisp code. a run-test.slosh script and one that test the slosh code with rust backed implementations (as opposed to slosh functions defined in slosh)
- working on lisp tests passing -> need to take detour to write read macro for lisp run-test.slohs script then I'll be off and running. ->  (would like to obviate the need to create dotimes-i in the float-bench.slosh script as well in that work)


vm/src/vm.rs
```diff
@@ -224,74 +224,150 @@ impl<ENV> GVm<ENV> {
...
+                (val1, Value::Value(h2)) => {
+                    let v2 = self.get_value(h2);
+                    val = self.is_equal_pair(val1, v2)?;
+                }
+                (Value::Value(v1), val2) => {
+                    let v1 = self.get_value(v1);
+                    val = self.is_equal_pair(v1, val2)?;
                 }
-                _ => {}
+                (_, _) => {} // TODO XXX: are these cases meaningful?
```
sstanfield, line 365: Yes, but the initial check (at the top) val1 == val2 will handle most of this.  For errors we might need to explicitly match since it is possible to create the "same" error as two different heap object.  For the other stuff I think it is very unlikely to have different heap objects for the same callable.  It may be possible (for instance compile same function twice) but even that case it may be ok to consider those different things...

slosh/src/docs.rs
```diff
@@ -360,21 +422,33 @@ impl SlFrom<SloshDoc> for Value {
...
+            Err(e) => Err(VMError::from(e)),
+        },
         _ => Err(VMError::new_vm("takes one argument (symbol)".to_string())),
     }
 }
 
+fn get_globals_sorted(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
+    if !registers.is_empty() {
+        return Err(VMError::new_vm(
+            "sizeof-value: takes no arguments".to_string(),
```
sstanfield, line 439: Fix the name here.

sstanfield (approved):

PR #129, opened 2024-02-28
Float bencher update

gpwclark:
@StevenLove even though the first two calls to eval-pol print out numbers in a given precision:
`399.999999999585` 
even though the correct answer as input by the program and correctly passed was: `399.99999999958527`

```
λ⸨⚙ ⸩ → ./bench.slosh
slosh 0.10.0 (test-new-floats:cee5608+, debug build, macos [aarch64], Feb 28 2024, 05:16:07 UTC [rustc 1.76.0 (07dca489a 2024-02-04)])
Using default slshrc written to "~/.config/slosh/init.slosh".
Edit this file to remove this message and customize your shell
(eval-pol 100 0.5) =>
399.999999999585
passed: true
(eval-pol 1000 0.05) =>
2105.263157892448
passed: true
(eval-pol 10000 0.2) =>
24999.999999998603
passed: true
```

is the printing working properly? Proof of issue is essentially 
1.  prints
```(eval-pol 100 0.5) =>
399.999999999585```
2 but
```(equal? (eval-pol 100 0.5) 399.999999999585)```
is false

### DIFF
--- a/slosh/benches/all.rs
+++ b/slosh/benches/all.rs
@@ -76,7 +76,7 @@ pub fn get_float_benchmark() -> PathBuf {
     get_benches_directory().join("float-bench.slosh")
 }
 
-fn run_float_script(n: usize, m: f32, expected: f32) {
+fn run_float_script(n: usize, m: f32, expected: f64) {
     let vm = &mut new_slosh_vm();
     let fname = get_float_benchmark();
     match std::fs::File::open(&fname) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by StevenLove** [comment]
I am still wrapping my head around the different layers. I'm able to load the eval-pol script into slosh and run it to see the behavior you described:
```bash
λ > (equal? (eval-pol 100 0.5) 399.999999999585)
false
λ > (equal? (eval-pol 100 0.5) 399.99999999958527)
true
```

I did some testing in Rust and found that the custom Display trait for F56 has a bug in it; it displays up to 12 digits after the decimal rather than up to 12 digits of precision total. Once fixed, those values with ~10 repeated nines should look much nicer.

PR #130, opened 2024-02-28
Add read and fix lisp tests

gpwclark:
add read, fix lisp tests so that they compile. Lots of errors but the syntax at least works.


slosh/src/load_eval.rs
```diff
@@ -207,7 +207,37 @@ fn eval(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
     }
 }
 
+fn read(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
```
sstanfield, line 210: The semantics of this are not quite write.  Probably want to call this "read-all" or something.  For the traditional read form it reads and returns one sexp at a time.  So if you used it with a string you would only get the first sexp and for a file you could loop over the file calling read to get all the sexps.  I know this is less useful for reading strings, hence the suggestion to change the name and reserve "read" for the "normal" lisp read (I will need to sort out string iterators soon that would make this useful with strings again).  This also means you always get a vec back which is also not the "normal" behavior but would make sense for a form that is intended to always read all the sexps it is given.

Anyway, this is a useful form but I don't think it should be given the name read...

gpwclark: @sstanfield fixed.

sstanfield (approved):

PR #133, opened 2024-03-01
Fix for F56::Display

StevenLove:


slosh/benches/bench.slosh
```diff
@@ -4,13 +4,13 @@
 
 (def pu (eval-pol 100 0.5))
 (prn "(eval-pol 100 0.5) =>\n" pu)
-(prn "passed: " (equal? pu 399.99999999958527))
+(prn "passed: " (equal? pu 400))
```
gpwclark, line 7: awesome!

slosh/benches/all.rs
```diff
@@ -92,7 +92,8 @@ fn run_float_script(n: usize, m: f32, expected: f64) {
     let last = run_reader(&mut reader);
     match last {
         Ok(Value::Float(f)) => {
-            assert_eq!(f64::from(f), expected);
+            assert_eq!(f.to_string(), expected.to_string());
```
gpwclark, line 95: remove this line and use the original. Since this is a benchmark we don't need to go through the extra cycles to check the string equality. You already made some good functions to do that.

Plus, we're keeping track of bench runs. And runs are kept track of and compared by name. So if we we change one it's important to rename it so you don't mix up benchmarks. Since the regular assert_eq works, and because this was more of a formatting issue when dealing with printed values in slosh (hence the fixes to bench.slosh) I think it makes sense to leave it as is. And I don't see any merit in making a new test that benchmarks some math in addition to the new to_string command. If you're representing a number as a string, there's going to be an alloc penalty anyway not to mention the fact that you're probably printing which is very expensive in terms of time. We probably should add some benchmarks for some typical actions of that sort, but we don't want those in these existing benchmark functions because I think it's a separate concern.

we do need to get bench.slosh running in CI now though.

vm/src/value.rs
```diff
@@ -152,17 +152,22 @@ impl Display for F56 {
...
+
+        // round to a max of F56::DIGITS sig figs
+        let orig_exponent_value = as_f64.abs().log10().floor() as i32; // the number after 'e' in scientific notation
+        let target_exponent_value = F56::DIGITS as i32 - 1; // exponent that we will shift this number to
+
+        let scale_factor = 10f64.powi(target_exponent_value - orig_exponent_value);
+        let scaled_and_rounded = (as_f64 * scale_factor).round();
+        write!(f, "{}", scaled_and_rounded / scale_factor)
+        // beware that 1e11 / 1e11 == 1, so this works
+        // but 1e11 * 1e-11 = 0.9999999999999 which would kill this algorithm
```
gpwclark, line 170: nice!

sstanfield (approved):

gpwclark (approved):

PR #136, opened 2024-03-01
F56 updates

StevenLove:
Extracted F56 to its own file with implementation and tests.
Also fixed an important bug from a typo in the F56::MIN_POSITIVE_SUBNORMAL
Focused tests on string conversion rather than conceptual equality to f64 values


vm/src/f56_module.rs
```diff
@@ -0,0 +1,581 @@
+use std::fmt::{Display, Formatter};
```
gpwclark, line 1: 1. probably fine to just rename this file f56.rs or float_56.rs. it >is< a module so the name is redundant.
2. Above this first line here put some "module" comments. Module style comments are:
//! This would describe a module at a higher level, explaining what it provides to the overall
//! functionality and is also a good place to discuss finer points or open questions in a project
//! like this.

when someone builds the docs (my personal favorite `cargo doc --open` opens in your default browser for you) module level comments are front and center when someone click on that module and it provides a nice high level overview to all the structs/fns/traits provided by that module/why it exists. We are going to have the docs automatically built and put on our github pages site once we're past this no tests business and it's next on the priority list.

gpwclark (approved):

vm/src/f56_module.rs
sstanfield: I agree with price on the name here, this is not a convention we want to introduce.

sstanfield (approved):
This is a good refactor.  Once the file f56_module losses the _module I think it can merge.

vm/src/f56_module.rs
StevenLove: Thanks for the feedback and explanations!
I've changed the module to float_56.rs and added some doc comments.

TIL `cargo doc --open`!

PR #138, opened 2024-03-02
Slosh test updates - need help!

gpwclark:
@sstanfield need help with defmacro, dec!, inc!, xar!, and xdr! once those are fixed the tests will all pass.

Also, nil I don' t think is behaving like a list, that's what getting in the way in xar! I believe. And in sl-sh calling first on a vec returned nil/the empty list, and now it's returning the empty vec. There are several places I had to change code to allow this new behavior. Is this intended?

`cargo test exec_global_slosh_tests_in_rust -- --nocapture` runs the doc's tests executed by rust.
in this test all 5 functions are failing: defmacro, dec!,inc!,xar!, and xdr! (excluding fs-exists? because I think it's shell calls won't work when run by cargo test)

`cargo test --features lisp-test --package slosh --test slosh-tests -- --nocapture` runs the run-tests.slosh script
in this test 4 functions are failing: dec!, inc!, xar!, and xdr!


builtins/src/string.rs
```diff
@@ -442,11 +442,11 @@ Make a new string by applying lambda to each char in input string.
 Section: string
 
 Example:
-(test::assert-equal "XstringXstrX" (str-map "xstringxstrx" (fn (ch) (if (= \x ch) \X ch))))
-(def test-str-map (str-map "xstringxstrx" (fn (ch) (if (equal? \x ch) \X ch))))
+(test::assert-equal "XstringXstrX" (str-map "xstringxstrx" (fn (ch) (if (equal? "x" ch) "X" ch))))
+(def test-str-map (str-map "xstringxstrx" (fn (ch) (if (equal? "x" ch) "X" ch))))
```
gpwclark, line 446: is there a syntax to get the characters x or X?

slosh/run-tests.slosh
```diff
@@ -1,26 +1,34 @@
...
+
 (let (globals (get-globals-sorted)
       globals-len (len globals)
-      i 0)
-    (load "init.slosh")
+      i 0
+      failing-tests #| {:dec! "dec!"
+                     :inc! "inc!"
+                     :xar! "xar!"
+                     :xdr! "xdr!"} |# )
```
gpwclark, line 11: this is how you avoid specific ones. I'll remove this before the merge.

vm/src/vm.rs
```diff
@@ -227,6 +227,15 @@ impl<ENV> GVm<ENV> {
             {
                 val = Value::True;
             }
+        } else if val1.is_nil() || val2.is_nil() || val1.is_undef() || val2.is_undef() {
```
gpwclark, line 230: should undefined be included here?

vm/src/vm.rs
```diff
@@ -366,6 +366,11 @@ impl<ENV> GVm<ENV> {
                     let v1 = self.get_value(v1);
                     val = self.is_equal_pair(v1, val2)?;
                 }
+                (Value::Nil | Value::Undefined, Value::True) => val = Value::False,
+                (Value::Nil | Value::Undefined, Value::False) => val = Value::True,
```
gpwclark, line 370: Should undefined be included here? What is it really?

lisp/core.slosh
```diff
@@ -435,12 +435,12 @@ Section: sequence
...
 (assert-equal '(2 3) (rest '(1 2 3)))
-(assert-equal '(2 3) (rest [1 2 3]))
+(assert-equal [2 3] (rest [1 2 3]))
 (assert-equal nil (rest '(1)))
-(assert-equal nil (rest [1]))
+(assert-equal [] (rest [1]))
 (assert-equal nil (rest '()))
 (assert-equal nil (rest nil))
-(assert-equal nil (rest []))
+(assert-equal [] (rest []))
```
gpwclark, line 443: 2 example of new behavior for list like data structures with vectors.

lisp/core.slosh
```diff
@@ -94,11 +94,12 @@ Example:
 (test::assert-true (seq? '(1 2 3)))
 (test::assert-true (seq? [1 2 3]))
 (test::assert-true (seq? '()))
+(test::assert-true (seq? '()))
 (test::assert-true (seq? []))
 (test::assert-false (seq? "aaa"))
 (test::assert-false (seq? 1))
 %#
-(def seq? (fn (v) (or (pair? v)(vec? v))))
+(def seq? (fn (v) (or (nil? v)(pair? v)(vec? v))))
```
gpwclark, line 102: changed function in lisp here but, should pair? recognize Nil?

slosh/src/docs.rs
```diff
@@ -532,10 +532,13 @@ mod test {
                         let symbol = d.symbol;
                         println!("{} ===============================", symbol);
                         println!("Should Run test for: {}", symbol);
+                        if symbol == "fs-exists?" {
```
gpwclark, line 535: this test was hanging  for me so I had to skip it.

slosh/run-tests.slosh
sstanfield: For dec! and inc! try removing the negative test with "xxx" as the second argument.  That is a compile error now so the entire test will fail to compile.  Not sure how the errors are working out but that is likely the issue, those usage sections simply won't compile in slosh (which is good since we are compiling inc!/dec! we know that second arg better be an int or at least possibly be an int...).

For xar!/xdr! it is no longer valid to use them on nil/().  That was a decision, supporting that adds complexity for a case I don't think is common (and is easily handled with fn if needed).  Open to discussion on this but for now you should remove usage that is not valid anymore.

vm/src/vm.rs
sstanfield: I included a description of Undefined in another PR, can discuss further if we need.

lisp/core.slosh
sstanfield: I think it sticking to the underlying type (vec in this case) and is expected.  We will need to do a review of all this after some dust settles.

lisp/core.slosh
sstanfield: I don't think so.  I know this could go either way but especially with xar!/xdr! needing a pair not nil I think it is best for it not to.  At least for now.  We need to have a big language discussion soon to hash out how we want some of this stuff to feel.  For instance, I am not sure having Nil/EmptyList may not make sense.

builtins/src/string.rs
sstanfield: Not sure I follow but "x" is not a char in slosh (or any lisp I know of).  It used to be \x and this was how you read char 'x'.

slosh/run-tests.slosh
gpwclark: I did, still not working. Here's a simple repro:

```
slosh> (def i 1)
1
Error getting prompt: [compile]: $sh: No such file or directory (os error 2)
slosh> (inc! i)
ERROR: [rt]: INC: Can only INC an integer type, got Special(Interned { id: 11 }).
prompt line: 42 ip: 0x00000004
DEBUG> :abort
```

gpwclark: xar! and xdr! are working though, and i fixed seq? so it no longer recognizes nil.

lisp/core.slosh
gpwclark: done.

slosh/run-tests.slosh
sstanfield: This is an a bug in inc/dec and globals.  I will push a patch once I sort it.

sstanfield (approved):
This is great and has already flushed a couple of things out.

PR #139, opened 2024-03-02
implementation of function to provide usage from db_args

gpwclark:
not sure how to iterate over arguments

### DIFF
--- a/lisp/core.slosh
+++ b/lisp/core.slosh
@@ -250,7 +250,13 @@ Print the documentation for provided symbol.
 
 Section: core
 %#
-(defmacro doc (sym) `(prn (eval (get-prop '~sym :doc-string))))
+(defmacro doc (sym)
+    (let (docs (eval `(get-prop ~sym :doc-string))
+         has-usage (str-starts-with "Usage:" (str-trim (str docs))))
+        (do
+        (if has-usage
+          (prn docs)
+          `(prn "Usage: " (eval (usage ~sym)) "\n\n" ~docs)))))
 
 #%
 Evaluate body a number of times equal to times' numerical value.
--- a/lisp/core.slosh
+++ b/lisp/core.slosh
@@ -254,9 +254,9 @@ Section: core
     (let (docs (eval `(get-prop ~sym :doc-string))
          has-usage (str-starts-with "Usage:" (str-trim (str docs))))
         (do
-        (if has-usage
-          (prn docs)
-          `(prn "Usage: " (eval (usage ~sym)) "\n\n" ~docs)))))
+        `(if ~has-usage
+          (prn ~docs)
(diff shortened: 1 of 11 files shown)


sstanfield (commented):
Lambdas and closures are primarily wrappers around a 'Chunk' which has a field dbg_args.  That should be an ordered list of the names that were given to the incoming arguments.  I think that will give you what you want.  If this does not make sense we can pair for a few when resurface to work this out.  Also, see debug.rs in slosh for some crude debug code that uses these, that might help.

gpwclark:
> Lambdas and closures are primarily wrappers around a 'Chunk' which has a field dbg_args. That should be an ordered list of the names that were given to the incoming arguments. I think that will give you what you want. If this does not make sense we can pair for a few when resurface to work this out. Also, see debug.rs in slosh for some crude debug code that uses these, that might help.

perfect that was very helpful. I think I got something that will work.

gpwclark:
@sstanfield 
final version of defmacro doc lookslike

```
(defmacro doc (sym)
    (let (docs `(get-prop ~sym :doc-string)
         has-usage (str-starts-with "Usage:" (str-trim (str docs))))
        (do
        `(if ~has-usage
          (prn ~docs)
          (prn "Usage: " (usage ~sym) "\n\n" ~docs)))))
```
worked on `./target/debug/slosh`

and I added (doc 'loop) and (doc 'inc!) to a script and it printed the right usage so I think we're good.

lisp/core.slosh
```diff
@@ -250,7 +250,13 @@ Print the documentation for provided symbol.
 
 Section: core
 %#
-(defmacro doc (sym) `(prn (eval (get-prop '~sym :doc-string))))
+(defmacro doc (sym)
+    (let (docs `(get-prop ~sym :doc-string)
+         has-usage (str-starts-with "Usage:" (str-trim (str docs))))
+        (do
+        `(if ~has-usage
```
sstanfield, line 257: I think this version is fine but it would be a little better to keep this if at compile/macro expand time and produce one or the other of the below prns (i.e. take out the ` here and add one to each prn branch).  This way the logic (that is redundant) to pick the correct prn does not get inserted into the runtime code.

As an aside, should the usage form be modified to always return a 'Usage', either the one in the docstring or the generated one?  I don't have an opinion on that but it seems like a potentially useful tweak?  Actually that may not be viable because the docstring is on the symbol so these questions may be misplaced.

sstanfield (approved):

lisp/core.slosh
gpwclark: taking both suggestions. I like keeping knowledge of the documentation structure in Rust. We know it meets the specifications at runtime off builds from master because the tests pass when we merge things.

PR #140, opened 2024-03-04
with lisp testing in place I can now write unit tests for rust code in the lisp std lib - back to macros

gpwclark:
Working on resurrecting the macro code so it works for slosh and the sl-sh std lib can be copied over with very little changes.

### DIFF
--- a/builtins/src/types/numbers.rs
+++ b/builtins/src/types/numbers.rs
@@ -4,6 +4,26 @@ use compile_state::state::SloshVm;
 use slvm::value::ValueType;
 use slvm::{from_i56, to_i56, VMError, VMResult, Value};
 
+impl SlFrom<()> for Value {
+    fn sl_from(_value: (), _vm: &mut SloshVm) -> VMResult<Self> {
+        Ok(Value::Nil)
+    }
+}
+
+impl SlFrom<&Value> for () {
+    fn sl_from(value: &Value, _vm: &mut SloshVm) -> VMResult<()> {
+        match value {
+            Value::Nil => Ok(()),
+            _ => Err(VMError::new_conversion(
+                ErrorStrings::fix_me_mismatched_type(
+                    <&'static str>::from(ValueType::Nil),
+                    value.display_type(_vm),
+                ),
+            )),
+        }
+    }
+}
+
 impl SlFrom<i32> for Value {
     fn sl_from(value: i32, _vm: &mut SloshVm) -> VMResult<Self> {
         Ok(to_i56(value as i64))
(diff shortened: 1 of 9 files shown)


builtins/src/types/numbers.rs
```diff
@@ -36,11 +56,53 @@ impl SlFrom<&Value> for i32 {
...
+}
+
 #[cfg(test)]
 mod tests {
     use crate::types::SlFrom;
     use crate::types::SlInto;
     use compile_state::state::new_slosh_vm;
+
+    #[test]
+    fn ts2() {
```
gpwclark, line 86: do not commit!

slosh/run-tests.slosh
```diff
@@ -5,10 +5,10 @@
 (let (globals (get-globals-sorted)
       globals-len (len globals)
       i 0
-      failing-tests {} #| {:dec! "dec!"
+      failing-tests {:dec! "dec!"
```
gpwclark, line 8: do not commit

slosh/src/docs.rs
```diff
@@ -532,7 +532,13 @@ mod test {
                         let symbol = d.symbol;
                         println!("{} ===============================", symbol);
                         println!("Should Run test for: {}", symbol);
-                        if symbol == "fs-exists?" {
+                        if symbol == "fs-exists?"
```
gpwclark, line 535: do not commit

builtins/src/types/primitives.rs
```diff
@@ -0,0 +1,83 @@
...
+use slvm::{VMResult, Value};
+
+impl SlFrom<&Value> for bool {
+    fn sl_from(value: &Value, _vm: &mut SloshVm) -> VMResult<Self> {
+        match value {
+            Value::True => Ok(true),
+            Value::False => Ok(false),
+            Value::Nil => Ok(false),
+            //TODO XXX should undefined be included here?
+            Value::Undefined => Ok(false),
```
gpwclark, line 12: Is undefined false?

sstanfield: Yes or it should panic.  Undefined is an internal state that the user should never see (may panic in cases if it surfaces- this indicates a compiler failure most likely).  It is used to initialize and clear the stack.  The real driver is that closed over values, if not cleared from the stack, can lead to random behavior when new code starts using random closed values...

The general idea is that properly compiled bytecode will always initialize a variable before accessing it and accessing an Undefined indicates this did not happen.  Of course "access" needs to constrained to something a bytecode did since it is part Value in some cases it is fine to access...

Ideally, without closed over values, we could just leave whatever junk was in a register there and replace as needed but we can't so this what I came up with.  I am open to simplifying this stuff and periodically think it.

sstanfield (approved):

PR #143, opened 2024-03-06
2. Refactor for macro import sanity

gpwclark:
functionally I just moved some things deeper into the library so I can refer to them from the builtins crate as I use the proc-macro. This brings us our first function str-sub.

basically in order to use macros you have to have fully qualified imports. Currently I'm using the macro in the builtins crate. but we could use it with any crate that imports compile_state and vm. Either way I had an issue where the macro was using the imports `builtins::add_builtin` for instance. And it would work in the tests trybuild tests, because those are their own binaries and compile against the public API and hence, see the builtins crate that way. But when I use the macro inside the builtins crate it exepcts the import to be `crate::add_builtins`. Anyway, let me know if we should put these elsewhere but it would be very handy if we could keep them in a dependency of builtins so I can both use and test them in there.


doc/containers.md
```diff
@@ -67,6 +67,8 @@ Other hashmap function:
 This is a traditional Lisp conscell data type, pair of values (car, cdr).  It can
 be used to create a linked list of values.
 
+Note: an empty list is not nil and nil is not an empty list.
```
gpwclark, line 70: this is true now right?

sstanfield: No, that is something I think maybe we should do.  Has not been implemented yet so if you have an opinion let me know.

sstanfield (commented):
I think I would rather see the stuff you moved into compile_state in it's own crate.  I don't really feel it belongs, what was you thinking in putting it here?  It does not look like you touched CompileEnvironment which kinda why compile_state exists (to create a SloshVM).

gpwclark:
> I think I would rather see the stuff you moved into compile_state in it's own crate. I don't really feel it belongs, what was you thinking in putting it here? It does not look like you touched CompileEnvironment which kinda why compile_state exists (to create a SloshVM).

totally get it. wasn't sure where to put them so I just put it in the minimal dependency. Created a new crate in the workspace. I think this is better.

compile_state/Cargo.toml
```diff
@@ -7,3 +7,4 @@ edition = "2021"
 
 [dependencies]
 slvm = { path = "../vm" }
+bridge_types = { workspace = true }
```
sstanfield, line 10: This dep does not seem to be required for compile_state.  Remove unless I am missing something...

bridge_adapters/src/lisp_adapters/string_char.rs
```diff
@@ -221,6 +220,16 @@ mod tests {
         val
     }
 
+    fn gensym(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
```
sstanfield, line 223: Why did this move to list_adapters?  This is a core lisp builtin so I don't see the logic here but maybe I am missing something.

sstanfield (approved):
Left a couple of small comments to address but otherwise I think this good.

bridge_adapters/src/lisp_adapters/string_char.rs
gpwclark: copy pasta sorry.

compile_state/Cargo.toml
gpwclark: removed.

doc/containers.md
gpwclark: It's on the radar because of the tests for sure.

PR #151, opened 2024-03-15
benchmark optimized

gpwclark:
move optimized code into benchmarking suite.


sstanfield (approved):
Running eval_pol 50,000 times on a CI box will probably be SLOW but if it is working maybe no big deal.

PR #152, opened 2024-03-16
a com-pilation-promise

gpwclark:
fortunately i'm running an arm computer. unfortunately it's still running the mac os.

@StevenLove  is also running mac, but with the intel arch.

and @sstanfield  i think you're on intel/linux.

since we can't easily get valgrind on mac, which is a requirement for iai-callgrind. I wanted to base it off os instead of architecture.

this does mean though @sstanfield that if you want to run the benchmarks that measure latency you'll have to mess with the cfg declaration in all.rs 

why would you ever need to run the criterion based tests. well. the `sl-py` repo that houses the python eval pol implementation (`pipenv run pytest --benchmark-json results.json benchmarks.py`) only runs latency tests. I don't see a python framework that supports the instruction/ram/l1 style enabled by valgrind.

that ok?

### DIFF
--- a/slosh/benches/all.rs
+++ b/slosh/benches/all.rs
@@ -152,7 +152,7 @@ fn run_continuation_search_script(n: usize) {
     run_bench_assert_true(n, run_continuation_search_bench, vm);
 }
 
-#[cfg(target_arch = "x86_64")]
+#[cfg(target_os = "linux")]
 mod instruction_count {
     use super::*;
     use iai_callgrind::{library_benchmark, library_benchmark_group, main, LibraryBenchmarkConfig};
@@ -225,7 +225,7 @@ mod instruction_count {
     }
 }
 
-#[cfg(not(target_arch = "x86_64"))]
+#[cfg(target_os = "macos")]
 mod wall_clock {
     use super::*;
     use criterion::{criterion_group, criterion_main, Criterion};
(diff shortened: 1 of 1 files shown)


sstanfield (approved):
I think this is fine.  Would it make sense to just do the wall clack run for all platforms (in main at bottom)?

gpwclark:
> I think this is fine. Would it make sense to just do the wall clack run for all platforms (in main at bottom)?

it definitely would. Until I started googling more I didn't think there was a good way to have the shared code that both bench frameworks use imported into two different files which is why I was using cfg annotations to effectively only have 1 main on 1 platform. but. I found this lovely SO answer: https://stackoverflow.com/questions/44539729/what-is-an-idiomatic-way-to-have-shared-utility-functions-for-integration-tests

1. I had NO idea about private shared crates. always forever. this is amazing.
2. It's super annoying to add a top level crate and I think it can get a little crazy and i've definitely added a lot of crates in the workspace.
3. I literally did to share the exec function in test_utils b/w crates and it's ugly.

1 works in this case. 
2 and 3 help when the shared code is really used in different crates so I don't think I can go back and fix my #[path] hack/annotation in slosh/main::tests

but maybe 1 CAN work b/w crates w/ relative imports? or if the workspace cargo.toml references the file... one second.

gpwclark:
> I think this is fine. Would it make sense to just do the wall clack run for all platforms (in main at bottom)?

so yeah wow. opened a new PR super happy with it. Didn't know private shared crates were a thing.

now my next complaint with bencher is  I wish I could add better branch names. they just show up as merge x or merge y. not as helpful.

PR #153, opened 2024-03-16
Fix adapter traits

gpwclark:
TLDR: intermediate update as I continue to add `sl_sh_fn` annotations to functions. 1 down and counting. Plus... i know it works because all the tests pass!

after our talk yesterday I realized certain traits needed to take mutable references to the slvm and certain traits didnt need to.  BridgedType is a marker trait that helps enforce but basically if you have a rust function like
```
fn allcaps_with_suffix(mystr: &str, suffix: &str) -> String {
    mystr.to_uppercase() + suffix
}
```
I will do something like
```
fn parse_allcaps(vm: &mut SloshVm, args: &[Value]) -> VmResult<Value> {
    let mystr: &str = args.get(0).sl_into_ref(vm);
    let suffix: &str = args.get(1).sl_into_ref(vm);
    {
        mystr.to_uppercase() + suffix
    }.sl_into(vm)
}
```
since sl_into_ref only needs an immutable reference to the SloshVm we're good and the mutable reference to allocate the new string at the end of the function is legal.

now:
-`SlFrom<SomeRustType> for Value` takes a mutable vm and allows you to take a rust type and return a Value, the mutable reference is important in case any allocations are needed.
- SlFromRef<Value> for SomeRustType takes an immutable vm and allow you to get references to the underlying rust types without doing any needless copying.
- I currently have SlAsMut and SlAsRef as well. Might get rid of them both and do SlFromRefMut<Value> for SomeRustType instead but we'll see. The semantics of both require that I return a reference where SlFromRef can it isn't required. I'll learn more as I keep researching.

current state:
I currently have one working bridge function `str-sub` and I'm making sure all the function signatures I expect to be able to compile can compile in `builtins/tests/` and that the errors for common things are what I expect.

- still need to get varargs working.
- still need to test turning a vector of a set of values (e.g. floats) into a `Vec<f64>`
- still need to experiment with tuple arguments
- if you have a function that returns one of it's arguments, mutable or no, I need a way to know inside the macro that the original value is being returned. as opposed to a rust type that would be created on the stack or allocated on the heap which would be the same type as the argument you passed in and intended to return... but it won't be *that* value.


vm/Cargo.toml
```diff
@@ -12,3 +12,4 @@ nohelmet = []
 
 [dependencies]
 unicode-segmentation = "1.10.1"
+bridge_types = { workspace = true }
```
sstanfield, line 15: This is cool but we will need to stay vigilant that bridge_types remains a lean no-dep crate.

sstanfield (approved):

vm/Cargo.toml
gpwclark: I promise!

PR #156, opened 2024-03-18
Floating Point Update

StevenLove:
Primarily, this PR allows us to change a single line to swap between F56 and F32Wrap implementations.

Secondarily, integration tests were changed so that they don't rely specifically on floats being 56 bit or 32 bit. This does come with a drawback. As long as we leave ourselves the option to switch between F56 and F32Wrap, an operation like  `(> 1.000000001 1.0))` is under-determined. With F56 it is true and with F32Wrap it is false.

Additionally (and this could be a separate PR), F56 is changed to only perform the expensive rounding operation when printing itself to a string or hashing. This comes from relaxing tests of floating point math to the same standards as f32 and f64.


vm/src/value.rs
```diff
@@ -247,11 +247,11 @@ impl Value {
         }
     }
 
-    pub fn get_float<ENV>(&self, _vm: &GVm<ENV>) -> VMResult<f64> {
+    pub fn get_float<ENV>(&self, _vm: &GVm<ENV>) -> VMResult<f32> {
```
sstanfield, line 250: Did you mean to drop this back to f32?  Doesn't this cause a lot of code to not even use the extra precision in f56?

sstanfield (approved):
I like this and it seems to get back to the original base line f56 performance.  I left one comment but pre-approving anyway.  I really like the ability to use the old f32 stuff for testing.  As a later PR we may want to make that a feature flag and possibly setup some benchmarks to quantify the cost of f56 on performance and/or allow a user to select low precision but higher performance when building (f56 should be the default).

StevenLove:
This PR is now being blocked by a clippy update that is now reporting problems with unrelated old code, so we will have to resolve those issues and then I can merge changes into this PR.

I can summarize the changes by saying I set the F56 PartialEq implementation to be a mostly-strict equality in that it considers +0 and -0 to be equal as well as considering every NaN to be equal (there are about a billion different bit-patterns that all represent NaN). I left in some unused implementations of other equalities 

I recommend you check out the doc I wrote contrasting the different forms of equality and how they get resolved. I found it very helpful to get a birds-eye view of the different equalities. 

Moving forward I would actually like to define within float_56 the different equalities to be used for `eq?` `equal?` and `=` such that `eq?` is the strict one and `equal?` and `=` are an identical looser function.

PR #158, opened 2024-03-22
More str macros

gpwclark:
now I'm just porting over functions from sl-sh.

I'm basically adding functions and making the changes necessary to re-support them as I go. 

The next hurdle is something like str-clear! where you want to accept a mutable reference to a string and return that same reference. I need a way for the user to be able to specify with an attribute which argument should be returned. or I need to do something with the lifetime. probably the former because the latter might be insane. I am parsing the generic parameter though, so I might be able to do something.


sstanfield (approved):
Responding to your comment about taking a returning a ref for something like str-clear!.  I don't think that will really work, I think taking a &mut String would be the proper way to receive a string that you were going to modify.  Not sure if you can even use a &mut str for some of this or how you would "glue" the in and out references or why you would want to (after all you have already modified the input ref so why return it).

PR #159, opened 2024-04-02
chore: appease clippy lint that ensure partial reads are handled

StevenLove:


StevenLove:
clippy lint update introduced some new issues.
I was able to resolve an error that related to a `read()` call in unix.rs.
But there's still a warning about `TypeBareFn` in bridge_macros lib.rs and I think one of you will have a much easier time fixing that issue so please take a look.

PR #162, opened 2024-04-03
1. testing new github pages stuff (initial)

gpwclark:
Update the github pages site:

[https://sl-sh-dev.github.io/sl-sh/](https://sl-sh-dev.github.io/sl-sh/)

1. it keeps the legacy docs in the sidebar for now so we can see what existed.
2. I threw in the rust docs b/c I could.
by going with [mdbook](https://rust-lang.github.io/mdBook/format/summary.html), I can use the crate as a library to create the md files with the slosh documentation.

mdbook is pretty extensible. You can write custom pre-processors, an existing one allows you to include the contents of other files, and you can write custom backends to output to things other than html. In theory it could be useful if we wanted to generate pdfs or something, but likely it could be useful with the md backend to generate md files dynamically.

Ultimately I would like to use the mdbook library to write a tool that will help us write documentation for the site. I would use this tool when iterating on an article. E.g. if I could write an article but specify some sort of test case or code block in slosh and then use a slosh pre-processor to write output from those commands rendered into specified areas in a md file.

Pretend this is a snippet from an article.
---------
Slosh offers one common command for getting the nth item of an object, `get`:
```preprocess
(doc 'get)
```
---------
with mdbook it would be easy to preprocess that (doc 'get) command and actually insert the results of the code. It would also be possible to fail if certain test cases failed. Perhaps those test cases could even be hidden in the rendered md file, but would fail to generate the md file if they did not pass.


slosh/Cargo.toml
```diff
@@ -22,6 +22,7 @@ glob = "0.3"
 # lisp-test dependencies
 regex = { workspace = true, optional = true }
 lazy_static = { workspace = true, optional = true }
+mdbook = { workspace = true, optional = true }
```
sstanfield, line 25: This is ok for a dev dep for building docs.  It is not OK for a normal dep, this thing pulls in everything and the kitchen sink (a http server, serde, tokio and lots of other stuff much with it's own large dep graph).  It looks like you are using it to make doc strings?  We need to find a better way for that if so, exploding the runtime dependencies for that is not worth the cost.

gpwclark: @sstanfield this is only compiling when `--features lisp-test` is true. That's why its an optional dependency and I amended this line in slosh/Cargo.toml to include this dependency:
`lisp-test = ["dep:regex", "dep:lazy_static", "dep:mdbook"]`

. I created this separate compilation target so that I could use external tools when building the docs that wouldn't cloud the actual runtime dependendencies of slosh.

gpwclark: To keep the functionality but further remove mdbook from slosh I could try and move the entire docs module into an examples directory? That way I could get a binary I could use to do things when building/testing docs (namely regex and mdbook... iirc what lazy_static does is now in the actual rust language more or less?).

I'd need a way to add some functions to slosh by incorporating the slosh crate as a depdendency though. Which is something we want to be able to do anyway I believe. But i'm not sure if it's possible now? Definitely willing to explore this route. 

When something is incorporated as a feature dependency versus a test dependency (as opposed to a default dependency) does it mess with the Cargo.lock / dependency resolutions mechanisms differently?

sstanfield: I missed that but I not super keen on a huge dep like that being one mistake away from being in slosh plus it adds friction to checking deps for slosh.  I think a separate app for building docs may be the way to go.  You could add a new crate to the workspace with the doc builder app if that makes more sense than an example.

As far as slosh being included as a crate, slosh was meant to be the top level "glue" that pulled in the various parts (vm, compiler, shell, etc) and built a binary.  So not really meant to be included, to do embedded slosh one would also import the various parts that were needed (for instance probably would always skip shell, might replace builtins or even the compiler etc).  If it is growing functionality that needs to be used elsewhere then that stuff maybe should live in a lib crate OR we may need a slosh_lib crate for this stuff and slosh will still just be a bin crate.

gpwclark: Ok this is good. I think this gives rise to a couple technical questions.

1. Is the approach I took to add in the "lisp-test" feature still ok if we wanted to say, add serde to slosh, and let users compile slosh with serde enabled to get some slosh builtin functions that could serialize/deserialize yaml/json/w/e? I do like the idea of having some opt in features but I understand the deps in those features shouldn't bring in everything and the kitchen sink.
2. In the case of mdbook I think something else is going on. I'm sort of building an enhanced version of slosh for a very specific purpose, it's very much a different crate sort of thing. Rust has a lot of libraries and slosh is a nice frontend for them, which is why I built the docs this way. However, I need a way to add those builtins to the SloshVm:
https://github.com/sl-sh-dev/sl-sh/blob/master/slosh/src/main.rs#L367

since main imports doc I can call the set_builtin function when SloshVm is created. If you could pass in something that implements a behavior that accepts the SloshVm at various times you could modify the behavior (e.g. I could add the docs builtins) but to do that I'd have to move almost everything in to slosh_lib and slosh would just pass in the noop behavior to slosh_lib for instantiating the SloshVm. 

Is that ok or do you have a better idea for dependency structuring to hook in the the SloshVm creation. The ultimate goal is to compile a different slosh binary that has additional builtin functions from rust.

gpwclark: Here's a possible implementation:

https://github.com/sl-sh-dev/sl-sh/pull/163

PR #164, opened 2024-04-07
1&2 + Custom preprocessor, automatically run slosh code in md files

gpwclark:
here are the two other previous PRs put together + the feature I wanted to use mdbook to build (if this isn't too gnarly this one can be approved and I'll close the other two, this PR does everything I'm trying to do for docs without adding deps to the slosh binary).

1. Slosh is much more extensible now. There is now a slosh_lib crate that slosh and slosh_test inherit from. This is pretty useful because it means someone else could pull in slosh_lib and add their own slosh functions w/ whatever dependencies they wanted.
2. slosh_test uses this extensibility and has a few additional dependencies (e.g. mdbook/regex) that are not in the regular slosh binary but needed for generating docs.
3. the doc testing stuff in github ci uses the slosh_test binary now.
4. the gh_pages.yml ci file uses the slosh_test binary now.
5. there is a new private crate in doc called mdbook-slosh-eval. It's marked as a workspace so it actually has its own Cargo.lock file and its dependencies are completely separate from the slosh binary as well.  This crate implements a preprocessor for mdbook that automatically reads and evaluates code blocks in md files that are fenced with the `slosh` tag (3 backticks followed by string `slosh` to start the code block instead of just 3 backticks) and writes the output to the bottom of the code block.

if this is acceptable it would supersede the other two.

By creating the private crate mdbook-slosh-eval in the doc/ directory and giving it access to slosh_lib, I am able to run slosh code when preprocessing markdown files with code blocks marked `slosh`.

To experiment on your own just run mk-site.sh in the ./doc subdirectory.

[The new README.md](https://github.com/sl-sh-dev/sl-sh/blob/22b78fe24f26dc4d3c3a030c6237d360214d7bf7/doc/README.md) in the ./doc subdirectory explains the functionality and I added it to several of the pages (the bit after the `=>` symbol comes from the pre-processor):
[new feature shown on let bindings md page](https://sl-sh-dev.github.io/sl-sh/let.html)
[new feature show on equality md page](https://sl-sh-dev.github.io/sl-sh/equality.html)


doc/src/equality.md
```diff
@@ -2,10 +2,18 @@
 
 When testing for equality in Slosh, there are several functions that can be used.
 Consider invocations like
-`(= 1 2)`,
-`(eq? 1 2)`,
-`(equal? 1 2)`, and
-`(assert-equal 1 2)`
+```slosh
```
gpwclark, line 5: to use the pre-processor you just need to append the string slosh to the appropriate code blocks

slosh_lib/src/shell_builtins.rs
gpwclark, line 1: Can't have a version string in a library so all the binaries have this code now instead.

sstanfield (approved):
This looks good.  I only skimmed the code that looked like it was just a move, if there was anything there you wanted reviewed that was actual changes let me know.

PR #165, opened 2024-04-10
More std lib

gpwclark:
- adding to the stdlib, a few questions peppered in.
- certain functions, e.g. reverse are slow and could be improved and don't need to make it in.


lisp/core.slosh
```diff
@@ -655,6 +673,87 @@ Section: core
...
+
+Section: core
+
+Example:
+(def test-apply-one (apply str "O" "NE"))
+(test::assert-equal "ONE" test-apply-one)
+(test::assert-equal 10 (apply + 1 2 7))
+%#
+(defmacro apply (a-fn & rst)
+    `(~a-fn ~@rst))
```
gpwclark, line 694: I couldn't get a version of apply to work that would return 6 for all of the following:

```slosh
(apply + 1 2 3)
(apply + (list 1 2 3))
(apply + 1 (list 2 3))
```
I could only get the first one working with this simple macro.

ideally it would work for all 3 right?

slosh_test/run-tests.slosh
```diff
@@ -9,11 +9,11 @@
     (while (< i globals-len)
         (prn "Symbol: " globals.~i)
         (prn "===")
-        (prn "doc: " (doc globals.~i))
         (def doc (doc-map globals.~i))
-        (prn "Usage: " (str doc.:usage))
-        (prn "Description: " (str doc.:description))
-        (prn "Section: " (str doc.:section))
+        ;; for some reason printing usage for vec-slice causes rust to panic with 'Handle 2045 is not a string!'
+        ;; (prn "Usage: " (str doc.:usage))
```
gpwclark, line 14: ran into an issue where if these were uncommented
`cargo test --workspace` would fail.

can you reproduce?

lisp/core.slosh
```diff
@@ -655,6 +673,87 @@ Section: core
...
+
+Example:
+(let (lst (list 1 2 3 4 5))
+    (test::assert-equal (list 1 2 2 4 5) (nsubstitute! lst 3 2))
+    (test::assert-equal (list 1 2 2 4 5) lst)
+    (test::assert-equal (list 1 3 2 4 5) (nsubstitute! lst 2 3 :first)))
+%#
+(defn nsubstitute!
+    (lst old-item new-item & mods)
+    (let (early-return (not (nil? mods)))
```
gpwclark, line 724: I wasn't able to make this check

`(early-return (in? mods :first)`

when I print out mods when called directly from nsubstitute! like:
```
(nsubstitute! lst 2 3 :first)
```
when the implementation starts like this:
```
(defn nsubstitute!
    (lst old-item new-item & mods)
    (prn "mods: " mods ", type: " (type mods))
    (let (early-return (in? mods :first))
    ...
```
it prints:
`mods: (#<Undefined>), type: :Pair`

when called directly running the shell it prints:
`mods: (:first), type: :Pair`

and I do not know why.

lisp/core.slosh
sstanfield: This is probably an edge case, the + "function" is compiled so it will not behave as expected when used like this.  This is something I will need to fix, I started some of the support for this a while ago but it is not hooked up (and I have not totally decided how to do it yet).  Obviously you need to be able to use "functions" that are compiled (in this case for speed since this becomes a single op code) need to work as expected when used.

slosh_test/run-tests.slosh
sstanfield: That error most likely means something was garbage collected that should not have been.  Sometimes this is caused by doing lisp things from Rust (i.e. allocating something and not making it "sticky" can cause this if it is not rooted).  At first glace this looks like it is all on the Lisp side though so may be a legit bug.  I will have to dig into this to try to figure out where the GC breakdown is.

builtins/src/collections.rs
```diff
@@ -70,58 +72,249 @@ pub fn vec_to_list(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
...
+
+pub fn flatten(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
+    let mut flat = Vec::with_capacity(registers.len());
+    flatten_helper(&mut flat, vm, registers)?;
+    Ok(vm.alloc_vector(flat))
+}
+
+/// Usage: (reverse items)
+///
+/// Produce a vector that is the reverse of items. Will call iter on items and requires a double ended iterator.
```
sstanfield, line 264: Double ended iter?  In the context of slosh this does not really mean anything, it works on collections and those should all be "double ended" not whatever we come up with iters in slosh.

Also, when we add back iters to slosh we may want this name for those (or maybe not), we can cross that bridge later.  This does make sense, just there may be overlap with future high level functionality.

sstanfield (approved):

PR #166, opened 2024-04-17
Fix the rest param issue and hopefully the invalid handle issue.

sstanfield:
@gpwclark This PR should fix the bug with nsubstitute! tests and hopefully the handle issue.  I am not seeing the handle issue so if you have a setup with the problem you may want to apply just the change to doc_map and see if it helps (pause/unpause GC while making the doc map).  I am confident I got it but sometimes there is more than one of these or something.

I removed builtins as a dep from compiler (makes things messier than I wanted) because I want to be able to compile from a builtin (circular dep otherwise).  Specifically I needed to fix dasm to work with any list for debugging this.  You may have opinions on how I did this, feel free to tweak in your PR if so.

Did not address your apply issue yet.


compiler/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2021"
 unicode-segmentation = "1.9"
 unicode_reader = "1"
 slvm = { workspace = true }
-builtins = { path = "../builtins" }
+#builtins = { path = "../builtins" }
```
gpwclark, line 12: remove this line

slosh_test/src/docs.rs
```diff
@@ -476,14 +476,22 @@ impl SlFrom<SloshDoc> for Value {
...
+            vm.pause_gc();
+            let res = match SloshDoc::new(*g, vm, Namespace::Global) {
+                Ok(slosh_doc) => Value::sl_from(slosh_doc, vm),
+                Err(DocError::ExemptFromProperDocString { symbol: _ }) => {
+                    let map = SloshDoc::nil_doc_map(vm);
+                    Ok(vm.alloc_map(map))
+                }
+                Err(e) => Err(VMError::from(e)),
+            };
+            // Unpause GC, this MUST happen so no early returns (looking at you ?).
```
gpwclark, line 491: lol. oh man. that's rough.

gpwclark (approved):

PR #167, opened 2024-04-21
Add an apply builtin.

sstanfield:
Add the lisp apply function.  It is partially optimized but could be better.


gpwclark (approved):
wow. that tail call. good find.

PR #169, opened 2024-05-06
Io builtins

sstanfield:
This adds back core IO, temp, rand and file meta data builtins.  The line count is obnoxious but most of this is pulling old slush builtins into slosh.  It is missing read-all extensions (to work with new IO objects) and read, these will come in there own PR- this one is large enough.  Also a bug in lisp hashmaps was uncovered in the fs-crawl tests so those are mostly commented out until fixed.  Basically a String and StringConst have different hashes which is unexpected in Lisp.  Will fix in its own PR as well.


gpwclark (approved):
w00t 

nice dude. this is a ton of good stuff!

PR #170, opened 2024-05-06
Implement clojure-style equality with =, not=, ==, and identical?

StevenLove:
Further information about this change is available in `doc/src/equality.md` file which explains the difference between the three core equality operations `=`, `==`, and `identical?`. They're also meant to map directly onto the same operations in Clojure. 

Original commits were squashed into 1, but their messages remain below
---
rename get_int macro to get_primitive_int
rename compare_int to compare_numeric
refactor NUM_LT, NUM_LTE, NUM_GT, NUM_GTE to use compare_numeric and removed helper macro
updated equality docs
implemented clojure-style = and == and not=
implemented identical

### DIFF
--- a/vm/src/vm/exec_loop.rs
+++ b/vm/src/vm/exec_loop.rs
@@ -805,7 +805,9 @@ impl<ENV> GVm<ENV> {
                 }
                 EQ => {
                     let (dest, reg1, reg2) = decode3!(self.ip_ptr, wide);
-                    let val = self.is_eq(reg1, reg2).map_err(|e| (e, chunk.clone()))?;
+                    let val = self
+                        .is_identical(reg1, reg2)
+                        .map_err(|e| (e, chunk.clone()))?;
                     set_register!(self, dest as usize, val);
                 }
                 EQUAL => {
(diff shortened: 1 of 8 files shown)


vm/src/vm/exec_loop.rs
```diff
@@ -948,30 +950,29 @@ impl<ENV> GVm<ENV> {
...
-                    true
-                ),
-                NUMLT => compare!(self, chunk, self.ip_ptr, |a, b| a < b, wide, true),
-                NUMLTE => compare!(self, chunk, self.ip_ptr, |a, b| a <= b, wide, true),
-                NUMGT => compare!(self, chunk, self.ip_ptr, |a, b| a > b, wide, true),
-                NUMGTE => compare!(self, chunk, self.ip_ptr, |a, b| a >= b, wide, true),
+                NUMEQ => {
+                    compare_numeric!(self, chunk, self.ip_ptr, |a, b| a == b, wide)
+                }
+                NUMNEQ => {
```
sstanfield, line 956: Good catch, not= should be (not (=... instead of (not (==... which I made it when I renamed operators.  In the old regime there was a numeric not equal op (/= I think) but now there is not.  Per other discussions, not= could be a macro or we could add a NEQUAL opcode that does this.  This opcode should be restored to numeric, you can leave it for this PR and I can clean this up in a bit.  Do we want a 'not==' for numeric not equal?  Could also be a macro, and maybe these should be, added the negative opcodes when we have NOT may be over-optimizing anyway...

sstanfield (approved):
We will need a follow up to fix the NUMNEQ mess unless you want to tackle that now.  I think I lean toward just moving the not= operator (and adding a not== op) to macros in core.slosh...  That opcode could be removed if we do but also OK to restore it to numeric not equal.  Will need to remove not= from the compiler as well if these become macros.  Either way I am OK with merging and addressing this later since your PR is correct from the user perspective and you added an issue to address this.

gpwclark (approved):
This is great. Second steve's question in the comments. maybe we can address that in the next sl-sh meeting?

also great docs, ty!!!

PR #172, opened 2024-05-10
Fix issues exposed by fs-crawl tests (map, inc!/dec! bugs).

sstanfield:
This is a bugfix PR, the fs-crawl tests exposed a couple of issue not related to fs-crawl...

- When using maps in lisp String and StringConst will hash to the same value for the same string.
- Fixed an issue with inc!/dec! on closed over values (they did not work).
- Tweaked the random tests so they should not fail about 1% of the time (see recent main CI failure).


gpwclark (approved):
mm, good catch with the hash map. I guess some of our types that wrap some of our types will run into this borrow checker game. 

i think it still make sense for the vm to be mutable though. the point isn't for the macro to do everything. it's for it to to make the things that are actually trivial, trivial.

PR #173, opened 2024-05-14
Add read

sstanfield:
This adds the read form and updates read-all for IO object (files).
- Improved examples for core file functions (proper temp file use and let over def).
- Improved (or at least simplified) read-line.
- The reader had to get hacked a bit to expose proper IO errors for the new read IO support.

### DIFF
--- a/builtins/src/io.rs
+++ b/builtins/src/io.rs
@@ -1,14 +1,12 @@
 use bridge_adapters::add_builtin;
 use compile_state::state::SloshVm;
 use slvm::{VMError, VMResult, Value};
-use std::borrow::Cow;
 use std::fs::{self, OpenOptions};
-use std::io::{Seek, SeekFrom, Write};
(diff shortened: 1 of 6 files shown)


gpwclark (approved):
wow. this was a bunch of work.

excited file io is this far along.
... aside

in sl-sh sometimes i felt like it was hard to get something from shell/process land into sl-sh land. mainly so that you could use sl-sh forms on say something from stdout.

now that file io exists...I assume treating stdout like a file and reading it into a string would be the easiest way to say, take stdout and run it through any slosh string form e.g. (str-contains (read *stdout*))?

I want to revitalize the regex forms, which obviously would work on sl-sh strings but it would be interesting if it were "no slower no" or not much slower than, just piping to grep.  My main motivation is the steve losh  batch coloring thing:
https://stevelosh.com/blog/2021/03/small-common-lisp-cli-programs/

Issue #171, opened 2024-05-06
Misleading opcode `NUMNEQ` does a non-numeric equality comparison


StevenLove:
We decided to mirror Clojure's equality design which creates a conflict due to us having implemented a numeric not equal function named `NUMNEQ` but Clojure only has a **non**-numeric not equal function more like `NOTEQUAL`.

I have made `not=` (and by extension `NUMNEQ`) behave like Clojure, so from slosh's perspective there is no way to notice that anything is off, but **internally** NUMNEQ is a misrepresentative name and it still goes through `compile_math.rs` despite being a non-numeric comparison. Due to the intentional ordering of the opcodes, I hesitate to just rename `NUMNEQ` to `NOTEQUAL` though that may be all that is required.

StevenLove:
`NUMNEQ` was removed from `opcodes.rs` in PR #176 which resolves this issue.

PR #176, opened 2024-05-23
Equality

StevenLove:
Finally merging equality changes. Check docs/equality.md for further details

### DIFF
--- a/doc/src/equality.md
+++ b/doc/src/equality.md
@@ -5,11 +5,14 @@ For numeric equality (IEEE) use `==`
 For bytewise equality use `identical?`
 
 The behavior and names are based on Clojure's implementation. Read their docs here: https://clojure.org/guides/equality
+Also check out `slosh/tests/equality.slosh` for some examples.
 
 -   `=`
+
     -   `(= 2 0x2)` is `true` (comparing an int to a byte)
     -   `(= 2 2.0)` is `false` (comparing an int to a float)
     -   `(= NaN NaN)` is `false`
+
     -   `state.rs` defines special forms object with key `equal` mapped to the name `=`.
     -   `compile.rs` matches on `env.specials().equal` and generates opcode `EQUAL`
     -   `exec_loop.rs` maps opcode `EQUAL` to function `is_equal`
@@ -18,6 +21,7 @@ The behavior and names are based on Clojure's implementation. Read their docs he
         -   check if both args are Byte or Int and if so, converts both to `i64` with `Value::get_int` and compares with rust native `==`
         -   check if both args are Byte or Int or Float and if so, converts both to `f64` with `Value::get_float`
             -   then subtracts the second from the first and takes the absolute value and tests `diff == 0.0`
+
 -   `==`
 
     -   `(== 1 1.0)` is `true` (comparing an int to a float)
@@ -37,7 +41,16 @@ The behavior and names are based on Clojure's implementation. Read their docs he
 
 -   `identical?`
 
-    -   uses `Value::PartialEq` implementation
+    -   `(identical? 1 1)` is `true`
+    -   `(identical? 1 1.0)` is `false`
+    -   `(identical? NaN NaN)` might be `true` or `false`. There are trillions of different bit patterns that represent NaN in IEEE 754
+
+    -   is the only equality comparison that uses `Value::PartialEq` implementation which is always false for different types of Values
+    -   using identical equality for floats causes problems with hashing.
+        [#125](https://github.com/sl-sh-dev/sl-sh/issues/125)
+        identical equality is 'too strict' in that you probably expect that +0 and -0 should hash to the same thing, but they don't
+        rendering hash tables
+
     -   `state.rs` defines special forms object with key `eq` mapped to the name `identical?`.
     -   `compile.rs` matches on `env.specials().eq` and generates opcode `EQ`
     -   `exec_loop.rs` maps opcode `EQ` to function `is_identical`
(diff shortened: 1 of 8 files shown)


StevenLove:
Apologies for the repeated PRs. I had to merge with master a few times because time had passed as well as needing to merge from a branch of sl-sh directly rather than from a branch on my fork in order to get bencher to work. 

The meat of the PR is the same as before, just a few commits to implement clojure-style `=`, `not=`, numeric `==`, and  `identical?`.  Still requires fixing issue #171 misleading opcode.

StevenLove:
added changes to resolve #171 by removing opcode implementation of NUMNEQ (not=) and instead using defmacro implementation in core.slosh

vm/src/vm/macros.rs
```diff
@@ -102,9 +102,9 @@ macro_rules! decode3 {
     }};
 }
 
-macro_rules! compare_int {
-    ($vm:expr, $chunk:expr, $code:expr, $comp_fn:expr,
-     $compf_fn:expr, $wide:expr, $move:expr, $not:expr) => {{
+macro_rules! compare_numeric {
+    ($vm:expr, $chunk:expr, $code:expr, $comp_fn:expr, $wide:expr) => {{
+        let should_move = true;
```
sstanfield, line 107: Why?  If $move was always true then get rid of this and collapse the if that uses it.  This was probably leftover from an earlier iteration.

vm/src/opcodes.rs
```diff
@@ -116,9 +116,9 @@ pub const DEC: OpCode = MATH_BASE + 5;
 // NUMEQ A B C - compare (=) in register B (inclusive) to C (inclusive) and set R[A] to the
 // result.
 pub const NUMEQ: OpCode = MATH_BASE + 6;
-// NUMNEQ A B C - compare (!=) in register B (inclusive) to C (inclusive) and set R[A] to the
-// result.
-pub const NUMNEQ: OpCode = MATH_BASE + 7;
+
+// NUMNEQ was here but it was removed and implemented as a macro instead
```
sstanfield, line 120: You don't need this comment.

vm/src/opcodes.rs
```diff
@@ -116,9 +116,9 @@ pub const DEC: OpCode = MATH_BASE + 5;
...
+
 // NUMLT A B C - compare (<) in register B (inclusive) to C (inclusive) and set R[A] to the
 // result.
 pub const NUMLT: OpCode = MATH_BASE + 8;
```
sstanfield, line 124: This should be +7 now not 8 since you removed NUMNEQ.

lisp/core.slosh
```diff
@@ -807,6 +807,40 @@ Example:
...
+
+Section: core
+
+Example:
+(assert-true (not= 0 1))
+(assert-true (not= 1 1.0))
+(assert-false (not= 2 2))
+(assert-false (not= 0.0 -0.0))
+%#
+(defmacro not= (expected-val right-val & body)
```
sstanfield, line 823: Get rid of "& body" it is not used.  This is adding a final parameter that will collect any extra params on a call, that is not what you want to do.  That would allow something like (not= 1 2 "blah" 4 8 6) to compile and be equivalent to (not= 1 2).  Macros often use this when they are wrapping an undetermined list of code.

lisp/core.slosh
```diff
@@ -807,6 +807,40 @@ Example:
...
+
+Section: core
+
+Example:
+(assert-true (not== 0 1))
+(assert-false (not== 1 1.0))
+(assert-false (not== 0.0 -0.0))
+(assert-false (not== 2 2))
+%#
+(defmacro not== (expected-val right-val & body)
```
sstanfield, line 840: Get rid of "& body".

vm/src/value.rs
```diff
@@ -245,6 +245,18 @@ impl Value {
         matches!(&self, Value::Byte(_) | Value::Int(_) | Value::Float(_))
     }
 
+    pub fn is_float(&self) -> bool {
+        matches!(self, Value::Float(_))
+    }
+
+    pub fn not(&self) -> Value {
+        match self {
+            Value::False => Value::True,
```
gpwclark, line 254: Change this to
```rust
Value::False | Value::Nil => Value::True
```
match statements in rust are super flexible. It's really cool.

On some timeline this can be a very cool read:
https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html

because there's a lot you can do.
It's also worth noting that the matches! macro essentially wraps an actual match statement so you can use all this funky syntax in there too.

vm/src/vm/macros.rs
StevenLove: Fair point; I was hesitant to remove `$move` completely as I didn't understand what it was meant to convey. I'll clean it up and update the PR.

lisp/core.slosh
StevenLove: thanks for the explanation!

doc/src/equality.md
```diff
@@ -1,55 +1,72 @@
...
+    -   based on `=`
+    -   is a macro defined in core.slosh which checks if the arguments are `=` and throws an error if they are not
+
+-   `not=`
+
+    -   defined in `vm/core.slosh` as the negation of `=`
+
+-   `not==`
+
+    -   defined in `vm/core.slosh` as the negation of `==`
```
gpwclark, line 72: great 
docs!

gpwclark (approved):

PR #181, opened 2024-06-06
Iterator foundation

sstanfield:
This also fixes a GC bug, asserts could double eval forms on failure and assert-error was broken.  Fixed some tests that only passed because assert-error was broken.

This is just the iterator start, needs a lot of stuff (filter, reduce, etc) but with bugfixes want to go ahead and get it in expand later.
The new iterators are a LOT simpler than in slush...


StevenLove (commented):
use 'params' instead of 'parms'

StevenLove (approved):

PR #183, opened 2024-06-09
Fix bug in for for destructuring, add enumerate, filter and reduce.

sstanfield:


lisp/iterator.slosh
```diff
@@ -246,8 +246,35 @@ Example:
     (mk-iter (if (< start end) (let (tmp start) (inc! start) tmp) :*iter-empty*)))
 
 #%
-Loops over each element in an iterator.  Will call iter on the input object.
-bind is bound to the current element of items and is accesible
+Iterator that wraps an iterator and generates pairs oF current index and value.
```
gpwclark, line 249:
```suggestion
Iterator that wraps an iterator and generates pairs of current index and value.
```

gpwclark (approved):

PR #186, opened 2024-07-02
Flesh out and document errors.

sstanfield:
This mostly documents the existing framework but is moving toward solidifying how errors work.  Also fixes a bug, could not use a continuation with on-raised-error for instance.


gpwclark (approved):

doc/src/errors.md
```diff
@@ -0,0 +1,27 @@
...
+(type [error]) returns :Error
+The form (car [error]) returns the keyword identifing this error.
+The form (cdr [error]) returns the data value for the error.
+Note the use of car/cdr to pull apart an error even though it is not of the pair type.
+Use (mk-err :[ID] value) to create an error type or (err :[ID] value) to "raise" an error (see below).
+
+## Raising an error
+Runtime errors will be "raised".  This means the program execution will halt and the debugger will be entered.  Currently this only allows examing the running state but will eventually include the ability to modify state and restart as other Lisps allow.  Code can raise an error with the err form, for instance (err :some-error-type "This is my error") will raise an error and interrupt the program.
+
+Note, the (get-error FORM+) form can be used to programmaticly return a raised error instead of breaking to the debugger.
```
gpwclark, line 16: So in theory. If I did not want any of my code dropping in to the debugger I would make the highest level form `get-error`? Just making sure I understand.

Either way this is very cool. I am going to want to build some stuff for the doc thing in lisp and i'm excited about the tools we're getting!

StevenLove (approved):

doc/src/errors.md
sstanfield: That is correct, if you add a top level get-error you should be able to not break into the debugger.

PR #187, opened 2024-07-11
Initial namespace support.

sstanfield:
See docs for load, run-script, ns, with-ns, import.

Add namespace support.  It is really just bookkeeping in the compiler but there were some interesting load issues.  The load special form now compiles the file at compile time (run-script works 100% at runtime like old load).

Also added a comp-time special form that runs arbitrary code at compile time (similar to macros).

Macros are installed fully at compile time now (using cost-time), this fixes a chicken egg macros issue when using them as you define them.

### DIFF
--- a/compile_state/src/state.rs
+++ b/compile_state/src/state.rs
@@ -202,11 +202,16 @@ pub struct Specials {
     pub is_ok: Interned,
     pub ret: Interned,
     pub ns: Interned,
+    pub with_ns: Interned,
     pub import: Interned,
+    pub load: Interned,
+    pub comp_time: Interned,
 
     pub rest: Interned,
     pub optional: Interned,
     pub scratch: Interned,
+    pub colon: Interned,
+    pub root: Interned,
 }
 
 impl Specials {
@@ -370,7 +375,6 @@ Add a sequence of numbers.  (+) will return 0.
 Section: math
 
 Example:
-(ns-import 'math)
 (test::assert-equal 0 (+))
 (test::assert-equal 5 (+ 5))
 (test::assert-equal 10 (+ 5 5))
@@ -385,7 +389,6 @@ Subtract a sequence of numbers.  Requires at least one number (negate if only on
 Section: math
 
 Example:
-(ns-import 'math)
 (test::assert-error (- 5 "2"))
 (test::assert-equal -5 (- 5))
 (test::assert-equal -5.0 (- 5.0))
@@ -401,7 +404,6 @@ Multiply a sequence of numbers.  (*) will return 1.
 Section: math
 
 Example:
-(ns-import 'math)
 (test::assert-equal 1 (*))
 (test::assert-equal 5 (* 5))
 (test::assert-equal 5 (* 1 5))
@@ -422,7 +424,6 @@ Divide a sequence of numbers.  Requires at least two numbers.
 
 Section: math
 Example:
-(ns-import 'math)
 (test::assert-equal 5 (/ 50 10))
 (test::assert-equal 5 (/ 50.0 10.0))
 (test::assert-equal 0 (/ 1 5))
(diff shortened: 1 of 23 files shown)


compile_state/src/state.rs
```diff
@@ -1071,10 +1074,111 @@ Example:
 (test::assert-true (ok? nil))
 "#),
             ret: add_special(vm, "return", ""),
+            ns: add_special(vm, "ns", r#"Usage: (ns SYMBOL)
+
+Changes to namespace.  This is "open-ended" change and is intended for use with
+the REPL prefer with-ns for scripts.
+THe symbol "::" will return to the "root" namespace (i.e. no namespace prepended to globals).
+This will cause all globals defined to have namespace:: prepended.
```
gpwclark, line 1082: all globals? or all non globals?

why would globals have `namespace::` prepended to them after using the `::` symbol?

compile_state/src/state.rs
```diff
@@ -1071,10 +1074,111 @@ Example:
 (test::assert-true (ok? nil))
 "#),
             ret: add_special(vm, "return", ""),
+            ns: add_special(vm, "ns", r#"Usage: (ns SYMBOL)
+
+Changes to namespace.  This is "open-ended" change and is intended for use with
+the REPL prefer with-ns for scripts.
+THe symbol "::" will return to the "root" namespace (i.e. no namespace prepended to globals).
```
gpwclark, line 1081:
```suggestion
The symbol "::" will return to the "root" namespace (i.e. no namespace prepended to globals).
```

compile_state/src/state.rs
```diff
@@ -1071,10 +1074,111 @@ Example:
...
+Example:
+(with-ns test-with-ns
+    (def ttf (fn () '(1 2 3)))
+    (test::assert-equal '(1 2 3) (ttf))
+    (test::assert-equal '(1 2 3) (test-out::ttf)))
+(test::assert-equal '(1 2 3) (test-out::ttf))
+"#),
+            import: add_special(vm, "import", r#"Usage: (import namespace [:as symbol])
+
+Will import a namespace.  Without an as then all symbols in the namespace will become available in the current
```
gpwclark, line 1112:
```suggestion
Will import a namespace.  Without an :as then all symbols in the namespace will become available in the current
```

compile_state/src/state.rs
```diff
@@ -1071,10 +1074,111 @@ Example:
...
+(test::assert-true x)
+(import testing :as t)
+(test::assert-true t::x)
+"#),
+            load: add_special(vm, "load", r#"Usage: (load path) -> [last form value]
+
+Read and eval a file (from path- a string).  The load special form executes at compile time.
+This means it's parameter must resolve at compile time.  Most of the time you will want to use
+this in conjuction with 'with-ns' to namespace the contents.
+Note: on it's own does noting with namespaces.
```
gpwclark, line 1133:
```suggestion
Note: on it's own does nothing with namespaces.
```

compile_state/src/state.rs
```diff
@@ -1306,6 +1458,56 @@ impl SloshVmTrait for SloshVm {
     }
 
     fn global_intern_slot(&self, symbol: Interned) -> Option<u32> {
+        fn check_global(vm: &SloshVm, ns: &str, sym: &str) -> Option<u32> {
+            let mut ns = ns.to_string();
+            ns.push_str("::");
```
gpwclark, line 1463: can namespaces be nested? e.g.

`reptar_lib::strings::misc`?

doc/src/namespaces.md
```diff
@@ -0,0 +1,50 @@
...
+(doc 'load)
+```
+
+### Run Script
+The run-script form loads each form in the file, compiles and executes it at runtime.  This means:
+- It can take any parameter since it is resolved at runtime.
+- Globals it defines will NOT be known until after it runs at runtime.
+
+```slosh
+(doc 'run-script)
```
gpwclark, line 49: really nice use of preprocessor! displaying the docs. what easy content haha.

gpwclark (approved):

compile_state/src/state.rs
sstanfield: Namespaces only apply to globals, locals live on the stack with limited scope so they would not matter.
Yes all globals defined in a namespace get namespace:: prepended.  This means you can not define something outside your namespace but you can "extend" it with embedded ::'s.  Lookups should be fine, it tries the prepended version as well as the non-prepended version.  Maybe allowing :: with def needs some more thought.

compile_state/src/state.rs
sstanfield: Not yet but I want to allow that.  It will require a variation on with-ns, currently with-ns saves the current ns, clears it and restores it when done.  Building a version that "extended" the current namespace would trivial and I will add that soon.

doc/src/namespaces.md
sstanfield: Yea, glad you pointed out that was there.

PR #188, opened 2024-07-11
Spellcheck

StevenLove:
I like the idea of spellchecking the codebase even though it's not entirely necessary. I'm using [CSpell](https://cspell.org/)

I'm able to hide away the relevant config files for this in a new root level .config/ directory where maybe one day some other ugly config can also go hide. Right now there are two files so it looks like this:

```
.config
    cspell.jsonc
    custom-dictionary.txt
```

I'm using the vscode extension so I see misspelled words highlighted as I'm typing and have a few more commands that I can use to easily add new words to the dictionary. Of course, adding that new word to the `custom_dictionary.txt` file is also not hard. It might work in other editors now or in the future (https://plugins.jetbrains.com/plugin/20676-cspell-check) (https://github.com/zed-industries/zed/issues/7553)

Alternatively, if you have an up-to-date nodejs installation, you can run `npx cspell .` to manually run spellcheck on your current working directory. I don't expect you to do this, I think the only productive way to go forward would be to get an extension for your editor running or to just manually check spelling issues every few months.


shell/src/platform/unix.rs
```diff
@@ -637,7 +637,7 @@ fn rlimit_to_c(rlimit: RLimit) -> Result<RlimitResource, io::Error> {
         #[cfg(not(any(target_os = "android", target_os = "linux")))]
         RLimit::SigPending => Err(Error::new(ErrorKind::Unsupported, "not on platform")),
         #[cfg(target_os = "freebsd")]
-        RLimit::KQueues => Ok(libc::RLIMIT_KQUEUESA),
+        RLimit::KQueues => Ok(libc::RLIMIT_KQUEUES),
```
StevenLove, line 640: please verify this is actually a typo

doc/src/errors.md
```diff
@@ -1,22 +1,22 @@
 # ERRORS
 
 ## Error Type
-The error type consists of an identifing keywork and a data/payload value.  This
```
StevenLove, line 4: should be identifying keyword*

shell/src/platform/unix.rs
sstanfield: Good catch, we may need to expand CI to other platforms.  From grepping libc the A is indeed a typo...

doc/src/errors.md
sstanfield: Yes.

sstanfield (approved):

gpwclark (approved):

gpwclark:
this is awesome.

PR #190, opened 2024-07-16
Remove slosh history line count from default init.slosh

StevenLove:
init.slosh has us printing the result of this statement `(" (str-trim $sh(cat ~/.local/share/sl-sh/history | grep -v "<ctx>" | wc -l)) ")")`

but it gives me `Reader error: Unquote outside of a back-quote` so we should remove or fix

### DIFF
--- a/init.slosh
+++ b/init.slosh
@@ -68,6 +68,6 @@
 (prn "		We don't have to be mean because,")
 (prn "		remember, no matter where you go,")
 (prn "		there you are.")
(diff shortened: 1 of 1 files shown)


sstanfield (approved):
Good catch, I have not been using the default recently so did not notice.  Not sure how this went totally unnoticed until now.  If you want to keep it this should work:
(prn (str "                     - Buckaroo Banzai (" ($sh #"%cat ~/.local/share/slosh/history | grep -v "<ctx>" | wc -l%")) ")")
Note the string literal syntax (#"%...%")- there is a bug in the reader that is not allowing escaped quotes in strings right now but the literal syntax works.  I will fix the escape quote soon as well then can use "...\"..\"..." syntax.

gpwclark (approved):

Issue #121, opened 2024-02-21
Installation on mac


StevenLove:
Following the docs for installation doesn't work for me on macOS 12.7.1.
Here is the relevant section

> ### 3. Use sl-sh as primary shell
> - install binary
> ```
> sudo install -D -m 755 target/release/slosh /usr/local/bin/
> ```
> - add slosh to /etc/shells and change login shell to slosh
> ```
> echo /usr/local/bin/slosh | sudo tee -a /etc/shells
> chsh -s /usr/local/bin/slosh
> ```

`install`
 - `sudo install -D -m 755 target/release/slosh /usr/local/bin/` doesn't work on my Mac
 - `install` on Mac is based on BSD and differs from GNU's `install` in that it doesn't support the `-D` flag.
 - We could modify the installation instructions to use `mkdir -p` instead.
 - Alternatively, we could have separate installation instructions for Mac

$PATH
 - On my machine, I need to add `/usr/bin/` to my $PATH for common utils like `which` and `less`. Should we modify the default `init.slosh` file to include `/usr/bin` in the $PATH?

 - We could also mention in these installation instructions for the user to consider manually migrating their `~/.bashrc` and `~/.bash_profile` to `~/.config/slosh/init.slosh`

gpwclark:
resolved by #191

PR #191, opened 2024-07-16
Update readme installation instructions for mac [skip ci]

StevenLove:
Addressing issue #121 makes these changes to the readme

- recommending that users performing a new installation should migrate their existing rc files' contents into their init.slosh file
- changing the `install` command from using the `-D` flag which is not supported on mac to doing the more universal `mkdir -p`
- adding /usr/bin to the PATH

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,8 @@ cargo run -p slosh
 - install binary
 
 ```
-sudo install -D -m 755 target/release/slosh /usr/local/bin/
+mkdir -p /usr/local/bin
+sudo install -m 755 target/release/slosh /usr/local/bin/
 ```
 
 - add slosh to /etc/shells and change login shell to slosh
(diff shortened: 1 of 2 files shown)


sstanfield (approved):

gpwclark (approved):

README.md
```diff
@@ -73,6 +74,11 @@ sudo install -D -m 755 target/release/slosh /usr/local/bin/
 echo /usr/local/bin/slosh | sudo tee -a /etc/shells
 chsh -s /usr/local/bin/slosh
 ```
+### 4. (Optional) Configure slosh
+The slosh configuration file lives at ~/.config/slosh/init.slosh. <br>
+If you run slosh and the file does not exist, a default one will be created for you. <br>
+Review your existing shell configuration files like ~/.bashrc and ~/.bash_profile and manually translate them to slosh syntax and add to your init.slosh file <br>
+For example, `export JAVA_HOME="/usr/local/opt/openjdk@11/bin/java"` becomes `(sh "export JAVA_HOME='/usr/local/opt/openjdk@11/bin/java'")`
```
gpwclark, line 81:
```suggestion
For example, `export JAVA_HOME="/usr/local/opt/openjdk@11/bin/java"` becomes `
(export 'JAVA_HOME "/usr/local/opt/openjdk@11/bin/java")`
```
export is a function in the slosh std lib as well so in an init.slosh

does the same thing but with less syntax and without invoking a new process with the `sh` command.

change the resultant slosh code to the line above.

PR #195, opened 2024-09-05
Fix Clippy issues from upgrading to Clippy 0.1.80

StevenLove:
apply clippy lint `for_loops_over_fallibles`
apply clippy lint `doc_lazy_continuation`
apply clippy lint `needless_borrows_for_generic_args `
ignore clippy lint `macro_metavars_in_unsafe`
fixed clippy lint `unexpected_cfgs` by removing code that would support a no_std build
remove unused feature options from vm

### DIFF
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -4,12 +4,6 @@ version = "0.1.0"
 authors = ["Steven Stanfield <stanfield@scarecrowtech.com>"]
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
-[features]
-gc = []
-nohelmet = []
-
 [dependencies]
 unicode-segmentation = "1.10.1"
 bridge_types = { workspace = true }
--- a/vm/src/vm/macros.rs
+++ b/vm/src/vm/macros.rs
@@ -1,7 +1,9 @@
 #[macro_export]
 macro_rules! inc_ip {
     ($code_ip:expr) => {{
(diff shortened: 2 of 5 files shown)


vm/Cargo.toml
```diff
@@ -4,12 +4,6 @@ version = "0.1.0"
 authors = ["Steven Stanfield <stanfield@scarecrowtech.com>"]
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
-[features]
```
sstanfield, line 9: I guess these are not used anymore?  I forget what gc was for since you kinda need it...  nohelmet was allow the use of more sketchy (but maybe faster) unsafe code.  Probably good if that one is gone.

vm/src/vm/macros.rs
```diff
@@ -1,7 +1,9 @@
 #[macro_export]
 macro_rules! inc_ip {
     ($code_ip:expr) => {{
+        #[allow(clippy::macro_metavars_in_unsafe)]
```
sstanfield, line 4: You can make an issue to revisit this.  What you did is fine but it does make me think I need to take a harder look at these macros...

sstanfield (approved):

vm/Cargo.toml
StevenLove: I came to the conclusion that they were entirely unused because I couldn't find any usage like an annotation `#[cfg(feature = "gc")]`. I agree you probably started with the garbage collection being optional and since made it mandatory. So no objection to removing these?

PR #196, opened 2024-09-05
Added justfile

StevenLove:
[Just](https://github.com/casey/just) is a popular command runner to replace `make`. I have struggled to remember and to type the exact wording of various tests so this allows me to type `just test` and cover all of my bases.  

Also note the script that checks the version of `cargo clippy` in the CI and compares it to your local `cargo clippy` version to remind you to update if necessary

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -8,8 +8,35 @@
 default:
     @just --list --unsorted
 
-# For all the recipes in this file, don't echo the commands to the terminal
-set quiet
+# Run all the necessary tests: clippy, unit tests, and integration tests
+test:
+    sh scripts/check-clippy-version.sh
+    cargo fmt
+    cargo clippy
+    cargo test --workspace
+
+# Generate and open the slosh docs
+sloshdoc:
+    cd doc && ./mk-site.sh
+    open doc/book/index.html
+
+# Generate and open the rust docs
+rustdoc:
+    cargo doc --workspace --no-deps --open
+
+
+# The remaining commands in the project are less commonly used
+
+# These tests are a subset of 'cargo test --workspace'
+# It runs ./slosh_test/tests/slosh-tests.rs which loads /slosh_test/run-tests.slosh which loads up all the globals and executes the Example block of each docstring
+# Test docstring examples of slosh globals defined with add_builtin or add_special
+test-globals:
+    cargo test --package slosh_test --test slosh-tests run_slosh_tests -- --exact --nocapture
+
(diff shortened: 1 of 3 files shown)


Justfile
```diff
@@ -0,0 +1,50 @@
...
+     cargo test --package slosh_test --test slosh-tests run_slosh_tests -- --exact --nocapture
+
+# These tests are a subset of 'cargo test --workspace'
+# Test slosh tests defined in /slosh/tests/*.slosh
+test-slosh-tests:
+    cargo test --package slosh --test lisp-scripts -- --nocapture
+
+# Use cspell to spellcheck the codebase
+spellcheck:
+    npx cspell .
```
gpwclark, line 50: how do i know i have installed everything I need to run just.

do i install just via cargo?
what is npx?

maybe add something in the README so we know how to approach. I like the idea of a centralized file that I can start sticking useful project stuff in, and the shell script to check the cargo version is quite clever!

sstanfield: I am fine with this, I don't want to required build deps beyond cargo but I don't think this does that.  Just supports another option plus records some handy command.

scripts/check-clippy-version.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
sstanfield, line 1: Would be nice if we ate our own dog food here.  I know there can be chicken/egg issues but we are building a shell so we should be able to write simple scripts with it.  Or fix it if we can not...

sstanfield (approved):

Justfile
StevenLove: Once you've [installed](https://github.com/casey/just?tab=readme-ov-file#packages) just, you can prove it by running `just -V` in your shell.

npx is another command-line utility, an alteration of npm the node package manager that comes with nodejs. Running `npx cspell .` downloads [cspell](https://www.npmjs.com/package/cspell) into a local cache on your machine and then runs it. And if it already existed in your local cache then of course it doesn't have to download it. This makes it quite simple to make use of nodejs scripts.

scripts/check-clippy-version.sh
StevenLove: Agreed!

Justfile
StevenLove: It's also possible to [set up shell completion scripts[(https://github.com/casey/just?tab=readme-ov-file#shell-completion-scripts) which makes it quite quick to see the potential commands

gpwclark (approved):

PR #200, opened 2024-09-19
feat: implemented form abs

StevenLove:
Also added new file for math builtins


sstanfield (approved):

StevenLove:
Thanks for approving. I should have merged and made a new PR for the `%` implementation but I didn't realize in time. 

---

I'd like some input on the family of remainder operations.

For background, there are perhaps 3 different valid remainder operation implementations, with their differences concerning how negative numbers are handled. For `(% 10 3) = 1`, the dividend is `10` and the divisor is `3`
- One's result is always a nonnegative number less than the divisor
- One's result has the same sign as the divisor
- One's result has the same sign as the dividend

clojure's [`rem`](https://clojuredocs.org/clojure.core/rem) and `mod`  implement two of these, and Rust's [`rem`](https://doc.rust-lang.org/std/primitive.i32.html#method.checked_rem) and `rem_euclid` implement two of these, but not the same two.

I've added the `%` and `rem_euclid` forms to mirror Rust's `rem` and `rem_euclid` but there's also an existing builtin in `builtins/lib.rs` that provides slosh with the `rem` form using this implementation that accesses registers. Is it written this way to make it more performant?
```rust
fn remainder(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
    let mut i = registers.iter();
    if let (Some(val1), Some(val2), None) = (i.next(), i.next(), i.next()) {
        let i1 = val1.get_int(vm)?;
        let i2 = val2.get_int(vm)?;
        Ok((i1 % i2).into())
    } else {
        Err(VMError::new_vm("requires two integers".to_string()))
    }
}
```

My question is: Should we follow Rust's example or Clojure's example? And can we simplify/unify our implementations by removing the old builtin implementation above?

gpwclark (approved):

Issue #202, opened 2024-09-21
Slosh docstrings like `(doc 'ns)` in `namespaces.md` cause errors and prevent `mdBooks` from building


StevenLove:
`namespaces.md` has five such errors preventing the script `./mk-site.sh` from completing
```bash
[2024-09-21T00:28:13Z DEBUG mdbook_slosh_eval::slosh_eval_lib] Evaluate slosh code block #4 in chapter: Namespaces
[2024-09-21T00:28:13Z DEBUG mdbook_slosh_eval::slosh_eval_lib] New Code Block: (doc 'ns)
    
    ;; => Encountered error: [compile]: Symbol usage not defined (maybe you need to use 'def usage' to pre-declare it).
```
This one caused by the source code:
````md
<!-- File: doc/src/namespaces.md -->
```slosh
  (doc 'ns)
```
````

After the 5 errors are reported, the thread panics:

```bash
thread 'main' panicked at slosh_test/src/docs.rs:677:16:
Building failed: Unable to parse the preprocessed book from "slosh-eval" processor
```

---

NOTE: `(doc 'ns)` works fine in `./target/debug/slosh`. However, perhaps the preprocessor doesn't have access to the functions (like `doc`) defined in `core.slosh`.

PR #203, opened 2024-09-24
fix mdbook address issue #202

gpwclark:
main change is when/how the usage function is loaded. Other than that just made some changes so that core/test were available to all the code loaded in `slosh` code blocks.


StevenLove (approved):
looks good!

doc/src/syntax-and-macros.md
```diff
@@ -0,0 +1,78 @@
...
+
+Expressions are enclosed in parentheses and Lisp uses prefix notation to parse expressions.
+A parenthesized expression is a list of forms: `(list 1 2 3 4)` where the first
+form, the head, should generally be a function or a macro and dictates what the expression will do,
+and the subsequent children are arguments to the head.
+
+## Comments
+
+Single-line comments start with a ; character and continue till the end of the line.
+Multi-line comments start with a #| and end with a |#
```
sstanfield, line 29: Also #; will consume and discard the next 'thing', usually an sexp- mostly for debugging.  Also could mention #%...%# for docstrings but those are special and not general comments.

doc/src/section-docs/hashmap.md
```diff
@@ -0,0 +1,33 @@
...
+(set! (get m "three") "changed3")
+(assert-equal "changed3" (get m "three"))
+```
+Clearing all keys is done with the `clear!` function.
+```slosh
+(def m {:one "one", :two "two", "three" 3})
+(clear! m)
+(assert-equal (make-hash) m)
+```
+** Hashmap keys that are not keyword types (not prefixed by a colon) can not be
```
sstanfield, line 30: This is a bug, you should (and one point) could do this.  At least I think you could.  At any rate I think you should be able to do this...

doc/src/section-docs/hashmap.md
```diff
@@ -0,0 +1,33 @@
...
+`set!`, and `clear!`.
+
+For retrieving items from a hashmap prefer the reader syntax (<object>.<keyword>) or use the `get`
+function.
+```slosh
+(def m {:one "one", :two "two", "three" 3})
+(assert-equal "one" m.:one)
+(assert-equal "two" m.:two)
+(assert-equal 3 (get m "three"))
+```
```
sstanfield, line 12: Might mention the ~symbol support here.  For instance '(def one :one) (pro m.~one) should work to allow the use of a variable after the dot (need this for looping with int keys for instance).

PR #207, opened 2024-10-03
fix: docs macro no longer double-printing usage string

StevenLove:
Problem 1: argument order was incorrect for `str-starts-with` so the result was always `false`.
Problem 2: some quoting and unquoting needed to be rearranged in order to get `has-usage` to actually be evaluated properly

Without this fix, you can issue a command like `(doc 'rem)` and see that the usage string is printed twice because the `doc` macro can't find an existing usage string and decides to try to add the usage string, doubling it.

### DIFF
--- a/lisp/core.slosh
+++ b/lisp/core.slosh
@@ -289,11 +289,11 @@ Section: core
 %#
 (defmacro doc (sym)
     (let (docs `(get-prop ~sym :doc-string)
-         has-usage (str-starts-with "Usage:" (str-trim (str docs))))
+         has-usage `(str-starts-with (str-trim (str ~docs)) "Usage:"))
         (do
-        (if has-usage
-          `(prn ~docs)
-          `(prn (str (usage ~sym) "\n\n" ~docs))))))
+        `(if ~has-usage
+          (prn ~docs)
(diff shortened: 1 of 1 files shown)


lisp/core.slosh
```diff
@@ -289,11 +289,11 @@ Section: core
 %#
 (defmacro doc (sym)
     (let (docs `(get-prop ~sym :doc-string)
-         has-usage (str-starts-with "Usage:" (str-trim (str docs))))
+         has-usage `(str-starts-with (str-trim (str ~docs)) "Usage:"))
```
sstanfield, line 292: This was the root issue, original version was always working from the sexp (get-prop SYM :doc-string) not the actual docstring.   Good catch.  I don't think you need the str around ~docs anymore, I think that was actually making the buggy call compile but the get-prop call that ~docs becomes should return a string now.

sstanfield (approved):
Good fix.  Couple of notes but I think your solution is fine.  The if thing does not really matter for how this macro is used and you could argue it would be more responsive to doc changes if used in code the way you wrote it.  You should try to get rid of the extra str though (noted), that should just be a wasted call now (harmless but useless).

PR #209, opened 2024-12-07
Update deps and fix new clippies.

sstanfield:
No functionality, just updates devs and fixes latest clippers.  Need to do this to sl-console and sl-liner as well.


gpwclark (approved):
duly noted. i'll randomly pick sl-liner to do, but if that's harder happy to switch.

sstanfield:
> duly noted. i'll randomly pick sl-liner to do, but if that's harder happy to switch.

@gpwclark 
Was just noting that we need to hit our other deps.  I don't think either will be an issue.  Also, either syn or quote (forget which) for the macros need a major bump and if I remember correctly the API has changed a bit.  If you want to tackle that first that would be good.  I left it for now, was just cleaning bitrot/taking care of some crates with security issues.

Also benches are broken, probably because an update so I have to look at that.

PR #218, opened 2025-02-04
finished migrating str->float yet

gpwclark:


bridge_adapters/src/lisp_adapters/numbers.rs
```diff
@@ -44,7 +131,7 @@ impl<'a> SlFromRef<'a, Value> for i32 {
     fn sl_from_ref(value: Value, vm: &'a SloshVm) -> VMResult<i32> {
         match value {
             Value::Int(num) => {
-                let num = from_i56(&num);
+                let num = I56::from_inner(&num);
```
gpwclark, line 133: I like the I56 thing just because it gives me something I can import that I can then perform I56 type things with. Don't care if we actually use it. More a comment on. There are some i56 types functions. i need them organized in such a way. we could also create a `mod i56 {}`?

bridge_types/src/lib.rs
```diff
@@ -123,6 +168,10 @@ impl ErrorStrings {
         }
     }
 
+    pub fn fix_me_invalid_string(expected: impl AsRef<str>, got: impl AsRef<str>) -> String {
```
gpwclark, line 171: still have no solution to how to do error messages... but I have a few ideas.

builtins/src/fs_meta.rs
```diff
@@ -48,9 +48,9 @@ fn cd_expand_all_dots(cd: String) -> String {
 ///
 /// Section: system
 ///
-/// Example:
-/// (def test-sleep-var (time (sleep 1000)))
-/// (assert-true (> test-sleep-var 1.0))
+// Example:
```
gpwclark, line 51: doing it in double quotes so it gets counted as a function that has no test.

builtins/src/string.rs
```diff
@@ -506,45 +507,52 @@ fn char_is_whitespace(target: SloshChar) -> VMResult<bool> {
...
+///  (test::assert-equal 101 (->int "101"))
+///  (test::assert-equal 1101 (->int :1101))
+///  (test::assert-equal 8 (->int 8.1))
+///  (test::assert-equal -111 (->int "-111"))
+///  (test::assert-equal 10 (->int "10.0"))
+///  (test::assert-error (->int "not int"))
+///  (test::assert-error (->int "--10"))
+/// ;; for some reason this ->sym test case must be last or else there's an error w/ load
+///  (test::assert-equal 1102 (->int (->sym 1102)))
+#[sl_sh_fn(fn_name = "->int")]
```
gpwclark, line 530: made the ->int and ->float variants you were talking about.

slosh_test_lib/src/docs.rs
```diff
@@ -604,7 +604,7 @@ fn build_each_docs_section_chapter(
         let file = format!("src/section-docs/{}.md", section);
         // If there is a section file header include it for preprocessing.
         if fs::metadata(&file).is_ok() {
-            content = content + &format!("{{{{ #include section-docs/{}.md }}}}\n\n\n", section);
+            content = content + &format!("{{{{#include src/section-docs/{}.md}}}}\n\n\n", section);
```
gpwclark, line 607: filed a bug. still dealing with this.

builtins/src/string.rs
```diff
@@ -506,45 +507,52 @@ fn char_is_whitespace(target: SloshChar) -> VMResult<bool> {
...
+///  (test::assert-equal 11.0 (->float 11))
+///  (test::assert-equal 10.0 (->float "10.0"))
+///  (test::assert-equal 13.0 (->float :13))
+///  (test::assert-equal 10.5 (->float "10.5"))
+///  (test::assert-equal 101.0 (->float "101"))
+///  (test::assert-equal -101.95 (->float "-101.95"))
+///  (test::assert-error (->float "not int"))
+///  (test::assert-error (->float "--10"))
+///  ;; for some reason this ->sym test case must be last or else there's an error w/ load
+///  (test::assert-equal 12.0 (->float (->sym 12)))
```
gpwclark, line 552: how do i specify a byte? I want to make sure my conversion code from a u8 to a [u8; 7] is correct. I assume I can just use the least significant bits from the byte and leave everything else as 0. But I don't know how to write a test case to create a byte @sstanfield

builtins/src/string.rs
```diff
@@ -506,45 +507,52 @@ fn char_is_whitespace(target: SloshChar) -> VMResult<bool> {
...
+///  (test::assert-equal 0 (->int "0"))
+///  (test::assert-equal 101 (->int "101"))
+///  (test::assert-equal 1101 (->int :1101))
+///  (test::assert-equal 8 (->int 8.1))
+///  (test::assert-equal -111 (->int "-111"))
+///  (test::assert-equal 10 (->int "10.0"))
+///  (test::assert-error (->int "not int"))
+///  (test::assert-error (->int "--10"))
+/// ;; for some reason this ->sym test case must be last or else there's an error w/ load
+///  (test::assert-equal 1102 (->int (->sym 1102)))
```
gpwclark, line 529: @sstanfield how do I create a Value::Byte? I want to test that i'm correctly modeling a u8 as a [u8; 7] (least significant bits replaced w/ source u8, everything else left as 0).

Cargo.toml
```diff
@@ -39,6 +39,7 @@ sl-compiler = { path = "compiler" }
 compiler_test_utils = { path = "compiler/test_utils" }
 slvm = { path = "vm" }
 shell = { path = "shell" }
+derive_more = {  version = "1.0.0", features = ["into", "from"]}
```
sstanfield, line 42: I am not going to die on this hill and I know derive_more is pretty standard.  But do we really need a new dep for these "Loose" types?  Seems like it save a few lines of code and adds more derive macro compile time.  Your call though, just my two cents.

sstanfield (approved):

doc/mk-site.sh
```diff
@@ -1,16 +1,22 @@
...
-git clone -b gh-pages-legacy-html https://github.com/sl-sh-dev/sl-sh src/legacy/
+git clone -b gh-pages-legacy-html https://github.com/sl-sh-dev/sl-sh legacy/ || true
 
 # don't want git submodules
-rm -rf src/legacy/.git
-mkdir src/rust-docs
-
-cargo doc --features lisp-test --target-dir src/all-rust-docs
-cargo doc --no-deps --document-private-items --features lisp-test --target-dir src/slosh-rust-docs
+rm -rf legacy/.git || true
```
gpwclark, line 9: TODO PC need to remove mk-site.sh and the build-doc function as things are now all done through the plugin.

doc/src/namespaces.md
```diff
@@ -7,12 +7,12 @@ Namespaces are compiler bookkeeping for organizing global symbols.  When in a na
 ## Entering a namespace
 
 From code use the 'with-ns' form.
-```slosh
+```sloshignore
```
gpwclark, line 10: @sstanfield i had to ignore these fields because they were breaking the build...

I believe what is happening is (doc) is writing to stdout and the extension communicates over stdin/stdout so when something else mucks with it then the mdbook can no longer be processed properly.

we wrote this code to get around that I think:
```
    fn exec_code(vm: &mut SloshVm, code: String) -> String {
        let code = format!(
            r#"(import test)
                (def *prn* "")
                (dyn
                    prn
                    (fn (&rest) (set! *prn* (str *prn* &rest)))
                    (do  {}))"#,
            code
        );

        let mut reader = Reader::from_string(code, vm, "", 1, 0);
        let s = slosh_test_lib::run_reader(&mut reader)
            .map(|x| x.display_value(&vm))
            .map_err(|e| format!("Encountered error: {}", e));
        match s {
            Ok(s) | Err(s) => s,
        }
    }
```
I'm a little stuck on how doc could still be writing to stdout given this is how we're executing the code. If you turn any of the ```sloshignore lines back into ```slosh lines you can reproduce the docs not generating by running `./mk-site.sh` when I'm done with this, all we will need to do is run `mdbook serve` and a webpage will be loaded locally and all changes to md files will be updated automatically.

PR #219, opened 2025-02-04
proposed fix, having trouble using curly braces in doc comments

gpwclark:
do we have string interpolation?

is that what i'm running into?

If we do (and I just broke it) the string interpolation needs to be backslash aware so we can still use curly braces in documentation comments right?


sstanfield (commented):
I don't think we need this plus we do want string interpolation in docstrings.  Escape does seem to work for me.  Me incoming PR has a fix to make 'doc' work with string interpolation as well.  Let me know if I am missing something after my next PR.

PR #222, opened 2025-02-09
Few new features and fixes.

sstanfield:
Added:
- 'get-namespaces' and 'get-in-namespace new forms.
- expose args as '*args*' (script exe is *shell-exe*)- first *arg* is the script in *args*.

Fixes:
- 'sh' should work in scripts now.
- Fixed a problem in the shell with export and strings.
- docstrings should work with interpolation now.
- $() will work on the REPL as expected now.
- Found and fixed another GC collection issue with real-all.


gpwclark (approved):
tested. looks great. and ty. this is exactly what I needed. Have a good iterative workflow on my branch now.

PR #223, opened 2025-02-16
Fixed a defer/eval interaction.

sstanfield:
Some state (defer list) was not being saved and restored in some cases.


gpwclark (approved):
nice catch!

Issue #224, opened 2025-02-16
only global namespace docs are built


gpwclark:
currently only docs from global namespace are built. new ability to pull in docs from all namespaces should make it possible to build and organize the rest of the documentation.

PR #227, opened 2025-04-20
Devops/fix gh pages mdbook

gpwclark:
- main fix was force installing mdbook in CI so the docs build correctly but I alos

1. edited some docs.
2. clarified some todos.
3. addresses issue #224 where non global namespaces weren't built
4. edited mdbook to support user documentation as that is a special case of the non-global namespace because there are slosh builtins that are not in the global namespace, e.g. iterator, that should appear in the std lib docs, as their namespace is importable. Any namespaces the user creates should not be mixed with the slosh section of the generated docs to avoid confusion and to help developers organize the symbols they create in their own section of the slosh docs.
5. updated some clippy errors
6. adapted a few more functions w/ the sl_sh_fn proc macro in hopes of making error messages consistent as per goals in #229.


README.md
```diff
@@ -14,12 +14,19 @@ scripting language, and it is a REPL.
...
 
-* The contains both a shell and a lisp reader so familiar bash-isms like
-    ```bash
+* The terminal contains both a shell and a lisp reader so familiar bash-isms like
+    ```
     cat file | tr -s " " | cut -d " " -f 2,4
     ```
-  "just work"
-* Support for an rc file, ```~/.config/slosh/init.slosh```, to set up environment and fully customize your prompt.
+  "just work", but starting a command with a lisp form (i.e. a balanced parenthesized s-expression) allows:
```
gpwclark, line 21:
```suggestion
  "just work", but starting a command with a lisp form (an s-expression) allows:
```

sstanfield (approved):

PR #232, opened 2025-05-07
allow any user file to be safely loaded within context of user doc generation

gpwclark:
new builtin functions to swap out builtin functions with the builtin noop function (returns Nil) and back. Used by the mdbook slosh eval crate to disallow writing to stdout by temporarily making pr/prn/dasm return Nil (noop).


gpwclark:
@sstanfield my one question was about how to set docs for global builtins. when they're instantiated it's just the string and the function. would like to start taking forms off the EXEMPTIONS map instead of tacking them on.

sstanfield (approved):

PR #233, opened 2025-05-16
User docs/fix2

gpwclark:
should have put this in earlier


slosh_test_lib/src/docs.rs
```diff
@@ -640,7 +640,17 @@ fn build_all_slosh_forms_listing_chapter(
     docs_by_section: &BTreeMap<String, Vec<SloshDoc>>,
 ) -> VMResult<Chapter> {
     let mut all_content = format!("# {}\n\n", name);
-
+    all_content += r#"
+When built locally (see doc/README.md) the mdbook generator has access to the user's
+slosh environment. This enables creating documentation in md form of all user defined
+functions. The user's init.slosh is automatically loaded, so any additional files it
+loads are alos imported. If the user needs to alter the slosh load path or add
```
gpwclark, line 647:
```suggestion
loads are also imported. If the user needs to alter the slosh load path or add
```

PR #234, opened 2025-06-15
new PR#2: std-lib/more2

gpwclark:
- add some more tests
- fix *load-path* to be tilde aware
- create some infrastructure to iterate over the builtin slosh files so it's easier to load all the namespaces necessary for doc stuff.
- some other refactoring


builtins/src/math.rs
```diff
@@ -136,7 +136,10 @@ pub fn rem_euclid(dividend: i64, divisor: i64) -> VMResult<i64> {
 /// Section: math
 ///
 /// Example:
-/// (test::assert-equal 1024.0 (2pow 10))
+/// ;; TODO PC asserting that this float assert-equal this int panics
```
gpwclark, line 139: @sstanfield we should talk about this.

vm/src/vm.rs
```diff
@@ -227,175 +227,169 @@ impl<ENV> GVm<ENV> {
...
+            }
+            (val1, Value::Value(h2)) => {
+                let v2 = self.get_value(h2);
+                val = self.is_equal_pair(val1, v2)?;
+            }
+            (Value::Value(v1), val2) => {
+                let v1 = self.get_value(v1);
+                val = self.is_equal_pair(v1, val2)?;
+            }
+            (Value::Nil | Value::Undefined, Value::True) => val = Value::False,
```
sstanfield, line 380: I am not sure undefined should be falsy.  If it sows up in this function we should probably panic, that most likely indicates a serious internal error.

vm/src/vm.rs
```diff
@@ -227,175 +227,169 @@ impl<ENV> GVm<ENV> {
...
     pub fn is_equal_pair(&self, val1: Value, val2: Value) -> VMResult<Value> {
         let mut val = Value::False;
-        if val1 == val2 {
-            val = Value::True;
-        } else if val1.is_int() && val2.is_int() {
-            if val1.get_int(self)? == val2.get_int(self)? {
-                val = Value::True;
+        match (val1, val2) {
+            (_, _) if val1 == val2 => val = Value::True,
+            (Value::Int(i1), Value::Int(i2)) if from_i56(&i1) == from_i56(&i2) => val = Value::True,
```
sstanfield, line 232: I think this is handled by (_, _).  Any of the basic (non-heap, non-float types) when comparing to the same type should be covered by (_, _) I think.

sstanfield (approved):

builtins/src/math.rs
gpwclark, line 139:
```suggestion
```

builtins/src/math.rs
```diff
@@ -136,7 +136,10 @@ pub fn rem_euclid(dividend: i64, divisor: i64) -> VMResult<i64> {
...
+/// ;; (test::assert-equal 1024 (2pow 10))
```
gpwclark, line 140:
```suggestion
```

PR #235, opened 2025-06-23
PR #2: std-lib/more (approved)

gpwclark:
I want to make sure bencher is running properly set with the proper thresholds so that we can be sure my changes to NUMEQ are ok.

- added a lot of new builtins
- added a way to easily run tests that aren't also in the documentation. Its not that I don't want all the tests documented. It's more that such an area should exist, because the feedback loops shouldn't always require rebuilding rust to load the Example: section of a docstring into the environment when we could just.. run slosh code.
- compare_numeric_eq is a modified version for comparing equality when both numbers are floats because otherwise users will run into really annoying errors at annoying times about precision that I don't think they care about.
- for all the math tests that want the numeric equality I added a new assert function that takes a comparison as well rather than change which equality function assert-equal uses. i pass in `==`.


slosh/tests/lisp-scripts.rs
```diff
@@ -34,48 +36,217 @@ pub fn get_globs_from_test_directory(glob: &str) -> Vec<PathBuf> {
...
+/// cargo test --package slosh --test lisp-scripts string/str-append.slosh
 /// ```
-/// cargo test --package slosh --test lisp-scripts -- --nocapture
-///```
 ///
-/// All slosh scripts in the ./slosh/tests/ directory will be run by this integration test. If
+/// All slosh scripts found will be run by this integration test. If
 /// the suffix of the file is _fail.slosh, then the test script is expected to fail. All other
 /// slosh scripts should return a 0 exit code or the integration test will fail.
 fn lisp_scripts() {
```
gpwclark, line 135: the main reason I made this was because I desperately needed shorter feedback loops. I like the tests in the documentation but I don't like not being able to target a given test or set of tests (rn we can only run all of them).

instead of making it s.t. you could tell the test framework to only run specific unit tests, I thought i'd make a more general purpose thing that can execute code in slosh files. esp. the ability to specify a directory and only run things from there.

builtins/src/collections.rs
```diff
@@ -366,48 +463,69 @@ Convert a vector to a list.
...
+        "vec-insert!",
+        vec_insert,
+        r#"Usage: (vec-insert! vector index new-element) -> vector
+
+Inserts new-element at index and moves following elements right in vector.  This is destructive!
+
+Section: vector
+
+Example:
+(def test-insert-nth-vec [1 2 3])
```
sstanfield, line 521: Just a note, this PR did not create this test.  We should start moving examples to more idiomatic (and figuring that out) slosh code.  Using def instead of a let binding for instance should be discouraged.  I did this a LOT but have been trying to clean up when working on code that intersects with docs.

vm/src/vm/exec_loop.rs
```diff
@@ -1048,7 +1048,7 @@ impl<ENV> GVm<ENV> {
                 MUL => binary_math!(self, chunk, self.ip_ptr, |a, b| a * b, wide),
                 DIV => div_math!(self, chunk, self.ip_ptr, wide),
                 NUMEQ => {
-                    compare_numeric!(self, chunk, self.ip_ptr, |a, b| a == b, wide)
+                    compare_numeric_eq!(self, chunk, self.ip_ptr, wide)
```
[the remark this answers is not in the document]
sstanfield: Yea, hopefully this does not slow down basic compares.  We probably need this anyway....

lisp/core.slosh
```diff
@@ -813,6 +918,3 @@ Example:
  %#
   (defmacro substitute (lst old-item new-item & mods)
        `(nsubstitute! (to-list ~lst) ~old-item ~new-item ~@mods))
-
-(load "iterator.slosh")
```
sstanfield, line 817: Not sure I love not loading iterator.slosh (test.slosh seems fine).  I guess it is not a bid deal to load it first but I kinda expect iterators to be part of core and always available, even if you have to prepend a namespace sometimes.

sstanfield (approved):
This is a bo one but lots of good stuff.

lisp/core.slosh
```diff
@@ -813,6 +918,3 @@ Example:
  %#
   (defmacro substitute (lst old-item new-item & mods)
        `(nsubstitute! (to-list ~lst) ~old-item ~new-item ~@mods))
```
gpwclark, line 920:
```suggestion
       `(nsubstitute! (to-list ~lst) ~old-item ~new-item ~@mods))
       
       (load "iterator.slosh")
```

PR #241, opened 2025-07-10
PR #1: Std lib/env

gpwclark:
- doc:
   show symbols that do not have documentation on the forms page so they are more discoverable.
- bencher:
  - add thresholds so we can be notified via gh actions if a metric is lagging
- std-lib:
  - fix
     some old scripts
  - add
     get-env
     set-env
    unset-env
    fs-fullpath

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,7 +25,7 @@ jobs:
       run: |
           cargo version
           rustc --version
-          cargo clippy --verbose -- -D warnings
+          cargo clippy --verbose -- -D warnings -A "clippy::uninlined_format_args"
     - name: Build
       run: |
           cargo version
--- a/builtins/src/fs_meta.rs
+++ b/builtins/src/fs_meta.rs
@@ -124,7 +124,7 @@ fn file_test(path: &str, test: fn(path: &Path) -> bool, fn_name: &str) -> VMResu
 
 /// Usage: (fs-fullpath "~/some/filepath/../") -> nil
 ///
-/// Return absolute `canonical` file path.
+/// Return absolute 'canonical' file path.
 ///
 /// Section: file
 ///
--- a/builtins/src/fs_meta.rs
+++ b/builtins/src/fs_meta.rs
@@ -122,6 +122,22 @@ fn file_test(path: &str, test: fn(path: &Path) -> bool, fn_name: &str) -> VMResu
     }
 }
 
+/// Usage: (fs-fullpath "~/some/filepath/../") -> nil
+///
+/// Return full filepath.
+///
+/// Section: file
+///
+/// Example:
+/// #t
+#[sl_sh_fn(fn_name = "fs-fullpath")]
+fn fs_fullpath(path: &str) -> Option<String> {
+    Path::new(path).canonicalize().ok().map(|x| {
+        let expanded = expand_tilde(x);
+        expanded.to_string_lossy().into_owned()
+    })
+}
+
 /// Usage: (fs-exists? path-to-test)
 ///
 /// Does the given path exist?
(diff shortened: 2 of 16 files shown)


.github/workflows/rust.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       run: |
           cargo version
           rustc --version
-          cargo clippy --verbose -- -D warnings
+          cargo clippy --verbose -- -D warnings -A "clippy::uninlined_format_args"
```
gpwclark, line 28: 🙄🙄🙄🙄🙄🙄🙄🙄🙄

if they leave this in i'm going to pitch a fit.

builtins/src/fs_meta.rs
```diff
@@ -122,6 +122,23 @@ fn file_test(path: &str, test: fn(path: &Path) -> bool, fn_name: &str) -> VMResu
     }
 }
 
+/// Usage: (fs-fullpath "~/some/filepath/../") -> nil
+///
+/// Return full filepath.
```
gpwclark, line 127:
```suggestion
/// Return absolute `canonical` file path.
```

compile_state/src/state.rs
```diff
@@ -480,9 +480,29 @@ Create a proper list from pairs with items 0 - N.
...
+            list_append: add_special(vm, "list-append", r#"Usage: (list-append list item)
+
+If second parameter is a list it will be appened to the first list, otherwise
+appends item to list.
+
+Section: pair
+
+Example:
+(test::assert-equal (list-append (list 1 2 3) (list 1)) (list 1 2 3 1))
+(test::assert-not-equal (list-append (list 1 2 3) (list 1)) (list-append (list 1 2 3) 1))
```
sstanfield, line 495: Maybe add a note or expand docs slightly to cover why these are not equal.  On first blush it looks like they would be...

shell/src/builtins.rs
```diff
@@ -332,6 +332,7 @@ where
 {
     let mut args = args.map(|v| v.resolve_arg(jobs).unwrap_or_default());
     let command_str = command.to_string_lossy();
+    // TODO sls how can we document builtins?
```
sstanfield, line 335: Good question.  I don't have a good answer yet.  We could add a symbol to attach docs to but that would get messy.  Maybe we need a shell builtin "doc" that covers shell builtins?

slosh_lib/src/shell_builtins.rs
```diff
@@ -151,45 +150,83 @@ fn sh_str(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
...
+/// environment variable and if it is it unsets it.
+///
+/// Section: core
+///
+/// Example:
+/// #t
+#[sl_sh_fn(fn_name = "unset-env")]
+fn unset_env(var: LooseString) {
+    unsafe {
+        env::remove_var(var.as_ref());
```
sstanfield, line 165: Need to scrub this input, see note on set_var.

slosh_lib/src/shell_builtins.rs
```diff
@@ -151,45 +150,83 @@ fn sh_str(vm: &mut SloshVm, registers: &[Value]) -> VMResult<Value> {
...
+/// value it should bind.
+///
+/// Section: core
+///
+/// Example:
+/// #t
+#[sl_sh_fn(fn_name = "set-env")]
+fn set_env(var: LooseString, value: LooseString) {
+    unsafe {
+        env::set_var(var.as_ref(), value.as_ref());
```
sstanfield, line 181: You need to scrub the input to set_var.  See the Rust docs and example in shell/builtins.rs (export and set_var functions), if you don't this can panic.

sstanfield (approved):
Aproving but you really do need to address the input scrub comments for env::{remove_var, set_var} for correctness.

